PR #4, opened 2019-05-31
Submit custom metrics via CloudWatch Logs

tianchu:
### What does this PR do?

Allow custom metrics to be submitted to CloudWatch Logs, and then be forwarded by Datadog log forwarder. This option can be enabled by setting environment variable `DATADOG_FLUSH_TO_LOG` to `True`.

### Motivation

Writting metrics to logs rather than sending to Datadog metric intake endpoint helps avoid adding latencies to the Lambda executions.

### Testing Guidelines

Upgrade to the latest version of datadog lambda layer and log forwarder.

### DIFF
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -5,7 +5,11 @@
 
 import os
 import sys
+import json
+import time
+import base64
 
+import boto3
 from datadog import api
 from datadog.threadstats import ThreadStats
 from datadog_lambda import __version__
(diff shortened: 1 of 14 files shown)


pinkerton:
Can you update the README with a sentence or two on how to use this?

datadog_lambda/metric.py
```diff
@@ -22,26 +24,42 @@ def _format_dd_lambda_layer_tag():
...
+    periodically and at the end of the function execution in a
+    background thread.
     """
-    _tag_dd_lambda_layer(args, kwargs)
-    lambda_stats.distribution(*args, **kwargs)
+    tags = _tag_dd_lambda_layer(tags)
+    if os.environ.get('DATADOG_FLUSH_TO_LOG') == 'True':
+        print(json.dumps({
+            'metric_name': metric_name,
+            'value': value,
```
DarcyRaynerDD: So, with the golang version of this, it would be simpler if I could send this as 'values:[value]', otherwise I'm going to have to write another serializer/marshaller. It also makes it compatible with the /series api endpoint. Is it a trivial change to do that, or does changing the log crawler function make it complicated?

tianchu: It's not a difficult code change to make, but it results in confusing code (e.g., `values: value`). Also, I'm calling this very `lambda_metric` function in log forwarder with the json writting to log, e.g., I can do simply do `lambda_metric(**metric_json)` with the current consistent signature.

BTW, while reading dd-lambda-go again, how are you preserving the actual `Timestamp` when the metric was submitted? I don't see `Timestamp` field in your struct for instance. Do they get timestamped when it's flushed? I

DarcyRaynerDD: No, the timestamping happens when the "Distribution" function is called. The "series" api takes a list of values as `values:[[$timestamp1, metricValue1], [$timestamp1, metricValue1]]`.

DarcyRaynerDD: My mistake, it uses "points" instead of values. Looks like it's incompatible enough that it doesn't make sense to try and make it match the API.
https://docs.datadoghq.com/api/?lang=bash#post-timeseries-points

DarcyRaynerDD: And, looking at it again, you're right, I think the timestamp isn't being saved. Damn, I'll go fix it.

PR #5, opened 2019-06-18
Shorten the metric log keys

tianchu:
### What does this PR do?

Use a slightly more compact format to log the submitted metrics.

### Motivation

Reduce CW log cost.


DarcyRaynerDD:
👍

PR #11, opened 2019-07-26
Support DD_LOG_LEVEL

tianchu:
### What does this PR do?

Allows setting datadog_lambda logger's loglevel via environment variable `DD_LOG_LEVEL` for ease of debugging.

### DIFF
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -8,11 +8,14 @@ import sys
 import json
 import time
 import base64
+import logging
 
 import boto3
 from datadog import api
 from datadog.threadstats import ThreadStats
 from datadog_lambda import __version__
+
+logger = logging.getLogger(__name__)
 
 lambda_stats = ThreadStats()
 lambda_stats.start()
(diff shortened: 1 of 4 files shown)


datadog_lambda/metric.py
```diff
@@ -52,13 +55,15 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None):
...
     if os.environ.get('DD_FLUSH_TO_LOG', '').lower() == 'true':
+        logger.debug('Sending metric to Datadog via log forwarder')
         print(json.dumps({
             'm': metric_name,
             'v': value,
             'e': timestamp or int(time.time()),
             't': tags
         }))
     else:
+        logger.debug('Sending metric to Datadog via lambda layer')
```
DarcyRaynerDD: Probably would be a good idea to print the metric name in this message. Also, if we could check that the API_KEY is set, or maybe print its length. I feel like people failing to set the API key correctly is the most common screw up, and something that lets us catch that would be helpful.

PR #14, opened 2019-08-14
Darcy.rayner/add trace intake

DarcyRaynerDD:
if you have not yet done so.*

### What does this PR do?

Adds ability to send trace metadata to trace intake


datadog_lambda/trace_intake.py
```diff
@@ -0,0 +1,89 @@
...
+                endTime=first_span.start + first_span.duration,
+            )
+        )
+        trace_payload = TracePayload(hostName=None, env="none", traces=traces)
+    return trace_payload
+
+
+def forward_traces(traces):
+    api_key, host = get_config()
+    conn = TraceConnection(host, api_key)
```
tianchu: Does this mean `forward_traces()` creates a new connection every time?

DarcyRaynerDD: Do we have to do anything special to keep a http connection warm in python? Or do they automatically get reused?

PR #18, opened 2019-10-03
Certain xray_recorder operations only work in the LambdaContext

tianchu:
### What does this PR do?
Certain xray_recorder operations only work in the `LambdaContext`, and it breaks when working in the non-lambda X-Ray `Context` (testing functions locally).

### Motivation

Fix https://github.com/DataDog/datadog-lambda-layer-python/issues/17

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -6,6 +6,7 @@
 import logging
 
 from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.lambda_launcher import LambdaContext
 
 from ddtrace import patch, tracer
 from datadog_lambda.constants import (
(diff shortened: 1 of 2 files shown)


datadog_lambda/tracing.py
```diff
@@ -154,3 +164,11 @@ def inject_correlation_ids():
...
 
     logger.debug('logs injection configured')
+
+
+def is_lambda_context():
+    """
+    Return True if the X-Ray context is `LambdaContext`, rather than the
+    regular `Context` (e.g., when testing lambda functions locally).
+    """
+    return type(xray_recorder.context) == LambdaContext
```
DarcyRaynerDD: Hmm, could this throw if X-Ray is not enabled? The technique I've been using to detect lambda is to look for the 'AWS_LAMBDA_FUNCTION_NAME' environment variable.

tianchu: I did consider checking the env var. But developers may have functions reading this env var (that's why it's provided by AWS), in this case, they may manually inject values to this env var for local testing as well. 

I also felt that env var is only a proxy for truth, and I got this idea from the xray sdk: https://github.com/aws/aws-xray-sdk-python/blob/e9d3746155729a28766f866714ce082eb6612ad3/aws_xray_sdk/ext/flask/middleware.py#L23

DarcyRaynerDD: Okay, makes sense.

PR #19, opened 2019-10-03
Submit metrics for invocations and errors

sfirrin:
### What does this PR do?

Submits aws.lambda.enhanced.invocations and aws.lambda.enhanced.errors metrics from the wrapper.

### Testing Guidelines

Added and updated unit tests. 

Also updated some of the functions in the demo account that the metrics are reporting for the functions with the new version of the layer.

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -7,16 +7,19 @@ import os
 import logging
 import traceback
 
-from datadog_lambda.metric import lambda_stats, lambda_metric
+from datadog_lambda.cold_start import set_cold_start
+from datadog_lambda.metric import (
+    lambda_stats,
+    submit_invocations_metric,
+    submit_errors_metric,
+)
 from datadog_lambda.patch import patch_all
-from datadog_lambda.tags import get_tags_from_context
 from datadog_lambda.tracing import (
     extract_dd_trace_context,
     set_correlation_ids,
     inject_correlation_ids,
 )
 
-ENHANCED_METRICS_NAMESPACE_PREFIX = "aws.lambda.enhanced"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


datadog_lambda/wrapper.py
```diff
@@ -40,19 +47,29 @@ class _LambdaDecorator(object):
...
+        logger.debug("datadog_lambda_wrapper initialized")
 
     def _before(self, event, context):
+        global cold_start_request_id
+        # Assign this request ID as the cold start if there is no value yet
+        if cold_start_request_id is None:
+            cold_start_request_id = context.aws_request_id
+
         try:
+            lambda_metric(
```
tianchu: Although this is a oneliner, I would make it a function in metric.py, to keep the wrapper focus on code flow rather than details.

datadog_lambda/wrapper.py
```diff
@@ -72,6 +89,13 @@ def __call__(self, event, context):
         self._before(event, context)
         try:
             return self.func(event, context)
+        except Exception:
+            lambda_metric(
```
tianchu: Although this is a oneliner, I would make it a function in metric.py, to keep the wrapper focus on code flow rather than details.

datadog_lambda/wrapper.py
```diff
@@ -40,19 +47,29 @@ class _LambdaDecorator(object):
...
         if self.logs_injection:
             inject_correlation_ids()
 
         # Patch HTTP clients to propagate Datadog trace context
         patch_all()
-        logger.debug('datadog_lambda_wrapper initialized')
+        logger.debug("datadog_lambda_wrapper initialized")
 
     def _before(self, event, context):
+        global cold_start_request_id
```
tianchu: I'm pretty sure `context.aws_request_id` stays the same for retried requests, which is gonna cause incorrect cold start?  Also if you are not using the actual request id, how about?

```python
# In a `cold_start.py` or `utils.py`?
initialized = False
cold_start = True

def set_cold_start():
    global initialized
    global cold_start
    cold_start = (initialized == False)
    initialized = True

def is_cold_start():
    global cold_start
    return cold_start

# In wrapper.py (here)
from utils import set_cold_start
    set_cold_start()

# use is_cold_start in, say, metric.py
from utils import is_cold_start
is_cold_start()
```

tianchu: Might be better to prefix the globals with `_` to mark them as "private".

PR #26, opened 2019-11-04
chore: bump dependencies

mbelang:
if you have not yet done so.*

### What does this PR do?

Update to latest dependencies

### Motivation

Our internal libraries wrapping ddtrace are using latest datadog library versions.

### Testing Guidelines

Running the tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,11 +28,11 @@ setup(
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4',
     install_requires=[
         'aws-xray-sdk==2.4.2',
-        'datadog==0.28.0',
-        'ddtrace==0.28.0',
+        'datadog==0.31.0',
+        'ddtrace==0.30.2',
         'wrapt==1.11.1',
-        'setuptools==40.8.0',
-        'boto3==1.9.160'
+        'setuptools==41.6.0',
+        'boto3==1.10.8'
     ],
     extras_require={
         'dev': [


setup.py
```diff
@@ -28,11 +28,11 @@
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4',
     install_requires=[
         'aws-xray-sdk==2.4.2',
-        'datadog==0.28.0',
-        'ddtrace==0.28.0',
+        'datadog==0.31.0',
```
hadrien:
```suggestion
        'datadog>=0.31.0',
```

hadrien: And so on?

mbelang:
@tianchu please.

mbelang:
Also what is the release procedure? Is it automated or I need to bump the version file manually with that PR?

tianchu:
@mbelang Hi Maxime, thank you for the pull request! 

For installations through AWS Layer, we do test everything before release, so it is safe to pick up the latest version for every release. However for installations through `pip` will automatically pick up the latest version whenever a lambda deploys. Breaking changes do occasionally & accidentally get released in minor/patch updates, and sometimes it takes days for the community to get things fixed. This also help us to ensure installations of the same version from AWS Layer and from pip share the same version of dependencies.

I can help test and publish a new version to AWS once you pin down the versions :)

mbelang:
@tianchu Sorry I didn't know that particularity with Lambda Layers. We install python dependencies through serverless framework and a python plugin. We use python poetry as a dependency manager which ensures versions via lock file. I understand it does make sense in that particular context though so I reverted to pinned versions.

Thank you

PR #27, opened 2019-11-05
Update dependencies

tianchu:
### What does this PR do?

Update dependencies

### Motivation

https://github.com/DataDog/datadog-lambda-layer-python/pull/26

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,17 +28,17 @@ setup(
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4',
     install_requires=[
         'aws-xray-sdk==2.4.2',
-        'datadog==0.28.0',
-        'ddtrace==0.28.0',
+        'datadog==0.32.0',
+        'ddtrace==0.30.2',
         'wrapt==1.11.1',
-        'setuptools==40.8.0',
+        'setuptools==41.6.0',
     ],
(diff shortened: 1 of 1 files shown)


mbelang:
Hi @tianchu. Do you have an estimate on this as it is a blocker for us right now. Than you

tianchu:
@mbelang Do you use pip or a requirements.txt to manage python dependencies? If so, you can list `ddtrace` with the desired version in your projects' requirements.txt directly, which would be override the `ddtrace` provided by the layer.

mbelang:
No i'm not. We are using poetry which does a better job a dependency resolution and there is obviously a conflict between our lib and this one as our lib uses the latest of datadog's

tianchu:
So you installs `datadog-lambda` using poetry? Does this workaround work for you https://github.com/sdispater/poetry/issues/697? 

You can also probably remove `datadog-lambda` from poetry, but instead use our serverless framework plugin: https://github.com/DataDog/serverless-plugin-datadog. Our plugin will install datadog-lambda as a AWS Lambda Layer instead of a python package, which would avoid causing conflicts with your own dependencies.

mbelang:
I decided to install datadog-lambda from git instead. I'm testing it ATM

tianchu:
The upgrade is blocked by a bug in the latest version (0.31.0) of `datadogpy`: https://github.com/DataDog/datadogpy/pull/480

PR #34, opened 2019-12-03
Allow kwargs in lambda handlers

SamuelM333:
### What does this PR do?

Allows `kwargs` in lambda handlers. 

### Motivation

Our serverless project has `kwargs` in the lambda handlers to make testing easier. If we add the datadog lambda wrapper, the tests break.  

Example of `kwargs` in a lambda handler:
```python
SQS = boto3.client('sqs', region_name='us-east-1')
DYNAMO_CLIENT = boto3.resource('dynamodb', region_name='us-east-1')
LOGGER = setup_logger()

def lambda_handler(event, context, logger=LOGGER, sqs_client=SQS, dynamodb_client=DYNAMO_CLIENT):
    pass
```
### Testing Guidelines

I ran `./scripts/run_tests.sh` and all tests were OK. I also installed locally this version of the package on my project and my tests now pass. 

### Additional Notes

I also modified the `run_tests.sh` shebang to `#!/bin/bash` since the `PYTHON_VERSIONS` array was declared with bash-specific syntax and trying to run `./scripts/run_tests.sh` with `sh` failed.

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -57,7 +57,7 @@ class _LambdaDecorator(object):
         patch_all()
         logger.debug("datadog_lambda_wrapper initialized")
 
-    def _before(self, event, context, **kwargs):
+    def _before(self, event, context):
         set_cold_start()
         try:
             submit_invocations_metric(context)
@@ -69,7 +69,7 @@ class _LambdaDecorator(object):
         except Exception:
             traceback.print_exc()
 
-    def _after(self, event, context, **kwargs):
+    def _after(self, event, context):
         try:
             if not self.flush_to_log:
                 lambda_stats.flush(float("inf"))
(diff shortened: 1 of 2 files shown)


datadog_lambda/wrapper.py
```diff
@@ -57,9 +57,8 @@ def __init__(self, func):
         patch_all()
         logger.debug("datadog_lambda_wrapper initialized")
 
-    def _before(self, event, context):
+    def _before(self, event, context, **kwargs):
```
tianchu: Is it necessary to pass `kwargs` to `_before()` and `_after()`? I don't think any of the datadog logic depends (or should depend) on the `kwargs`.

SamuelM333: Makes sense. Gonna remove that.

tianchu:
@SamuelM333 The PR is now merged and will be included in the next release. Thanks for your contribution 👍

PR #41, opened 2020-01-15
check api_key had been set before setting the value from env

neong83:
if you have not yet done so.*

### What does this PR do?

Lambda metric will tries to get `DD_KEY_API` value from 3 different places when `lambda_metric` imports and set its value into `api._api_key` in Datadog library.  However, I received an empty string as result when I tries to import `lambda_metric` that describe in doc
```
https://github.com/DataDog/datadog-lambda-layer-python#environment-variables 
from datadog import api
api._api_key = "MY_API_KEY"
```

This PR is intend to check on `_api_key` before setting it.

### Motivation

allow user to overwrite API Key at runtime without side effect

### Testing Guidelines

- [ ] `_api_key` will preserve when `Datadog.lambda_metric` imports if we had previously set it in `Datadog.api._api_key` 

### Additional Notes

N/A

### DIFF
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -108,19 +108,20 @@ def submit_errors_metric(lambda_context):
 
 
 # Set API Key and Host in the module, so they only set once per container
-DD_API_KEY_SECRET_ARN = os.environ.get("DD_API_KEY_SECRET_ARN", "")
-DD_KMS_API_KEY = os.environ.get("DD_KMS_API_KEY", "")
-DD_API_KEY = os.environ.get("DD_API_KEY", os.environ.get("DATADOG_API_KEY", ""))
-if DD_API_KEY_SECRET_ARN:
-    api._api_key = boto3.client("secretsmanager").get_secret_value(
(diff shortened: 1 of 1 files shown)


tianchu:
@neong83 do you override api key after importing datadog_lambda?

```python
from datadog_lambda.metric import lambda_metric

# Override DD API Key after importing datadog_lambda packages
from datadog import api
api._api_key = "MY_API_KEY"
```

neong83:
@tianchu I did have to override the api key by the way you posted in your comment. 

The reason being I was able to set api key directly into datadog.api._api_key inside django.settings and  `datadog_lambda_wrapper` will use it throughout the entire code base.  But I would have to reset the key whenever I import `lambda_metric` from `datadog_lambda.metric`.  It is causing unexpected behavior for developer and have to remember resetting it when this pattern is used.

It does fix the problem if I set the api key as an environment variable, but this is causing some unnecssary overhead of reloading / resetting the key every time lambda metric gets import.

Hope this answers your questions and concerns

PR #46, opened 2020-03-06
Fix duplicate enhanced metrics

tianchu:
### What does this PR do?

If the datadog serverless plugin is used, which automatically wraps the function's handler with `@datadog_lambda_wrapper`, and the handler function is also decorated with `@datadog_lambda_wrapper`, the resulted code is equivalent to the following, and two wrappers get created. 

```python
@datadog_lambda_wrapper
def lambda_handler(event, context):
    pass

def plugin_generated_lambda_handler(event, context):
    datadog_lambda_wrapper(lambda_handler)(event, context)
```

This results in `__init__` get called twice on cold start and `__call__` gets called twice on every invocation. Because not every operation in those two methods are idempotent, this causes unexpected behaviors. E.g., `aws.lambda.enhanced.invocations` will get submitted twice for every single invocation.

Let `@datadog_lambda_wrapper` instantiate a no-op decorator `_NoopDecorator` when a wrapper is already instantiated. But allow `_force_new` to force create a new/real wrapper every time, which is needed by the unit tests.

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -38,13 +38,39 @@ def my_lambda_handle(event, context):
 """
 
 
+class _NoopDecorator(object):
+
+    def __init__(self, func):
+        self.func = func
+
+    def __call__(self, *args, **kwargs):
+        return self.func(*args, **kwargs)
+
+
 class _LambdaDecorator(object):
     """
     Decorator to automatically initialize Datadog API client, flush metrics,
     and extracts/injects trace context.
     """
 
+    _instance = None
+    _force_new = False
+
+    def __new__(cls, func):
+        """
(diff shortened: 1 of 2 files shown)


datadog_lambda/wrapper.py
```diff
@@ -38,13 +38,39 @@ def my_lambda_handle(event, context):
...
+        If the decorator is accidentally applied to multiple functions or
+        the same function multiple times, only the first one takes effect.
+
+        If _force_new, always return a real decorator, useful for unit tests.
+        """
+        if cls._force_new or cls._instance is None:
+            cls._instance = super(_LambdaDecorator, cls).__new__(cls)
+            return cls._instance
+        else:
+            return _NoopDecorator(func)
```
sfirrin: nice idea

PR #48, opened 2020-03-10
Add integration tests

sfirrin:
### What does this PR do?

These integration tests do the following:

1. Build layers using the ./scripts/build_layers.sh script (or don't if $BUILD_LAYERS isn't set)
1. Deploy sample Lambda handlers for all runtimes using the serverless framework
1. Invoke each Lambda once for each of the event types defined in ./tests/integration/input_events
1. Compare the return value of each Lambda invocation for each input event with the snapshot that we have saved for the input event in ./tests/integration/snapshots/return_values
1. Query for the logs of each Lambda function
1. Replace invocation-specific data (like timestamps, request IDs) with XXXX to normalize log snapshots
1. Compare new logs to the snapshot log file that we have saved in ./tests/integration/snapshots/logs
1. Print any mismatches between snapshots and new logs / return values
1. Fail the script if a mismatch is found, otherwise succeed

To run integration tests for new changes to the layer:
`BUILD_LAYERS=true DD_API_KEY=XXXX aws-vault exec sandbox-account-admin -- ./scripts/run_integration_tests`

To regenerate snapshots:
`UPDATE_SNAPSHOTS=true DD_API_KEY=XXXX aws-vault exec sandbox-account-admin -- ./scripts/run_integration_tests`

This PR also makes a change to the layer itself to hook into the existing request patching so that if env var `DD_INTEGRATION_TEST=true`, each HTTP request will be logged in this format:
`HTTP GET https://ip-ranges.datadoghq.com/ Headers: ["x-datadog-parent-id:XXXX", "x-datadog-sampling-priority:2", "x-datadog-trace-id:XXXX"] Data: {}`
so that we can see in the snapshot tests if any HTTP requests have changed. 

This is the first step of the integration tests so we are only running them locally for now. In the future we will integrate these tests into our CI.

### Testing Guidelines

I've run this many times locally and it is working as expected without being flaky.

### Additional Notes


scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,157 @@
...
+    for runtime in "${RUNTIMES[@]}"; do
+        function_name="${handler_name}_${runtime}"
+        # Invoke function once for each input event
+        for input_event_file in "${input_event_files[@]}"; do
+            # Get event name without trailing ".json" so we can build the snapshot file name
+            input_event_name=$(echo "$input_event_file" | sed "s/.json//")
+            # Return value snapshot file format is snapshots/return_values/{handler}_{runtime}_{input-event}
+            snapshot_path="./snapshots/return_values/${function_name}_${input_event_name}.json"
+
+            return_value=$(serverless invoke -f $function_name --path "./input_events/$input_event_file")
```
DarcyRaynerDD: The invoke with return value is a neat trick

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,157 @@
...
+for handler_name in "${LAMBDA_HANDLERS[@]}"; do
+    for runtime in "${RUNTIMES[@]}"; do
+        function_name="${handler_name}_${runtime}"
+        function_snapshot_path="./snapshots/logs/$function_name.log"
+
+        # Fetch logs with serverless cli
+        raw_logs=$(serverless logs -f $function_name --startTime $script_start_time)
+
+        # Replace invocation-specific data like timestamps and IDs with XXXX to normalize logs across executions
+        logs=$(
```
DarcyRaynerDD: Also pretty cool trick

PR #50, opened 2020-03-12
Try never break the wrapped lambda function

tianchu:
### What does this PR do?

- patch `requests` only when it's installed and used
- do not throw when `requests` is called explicitly with `headers=None` (should be a dict)
- swallow exceptions from datadog wrapper logic

### Checklist

### DIFF
--- a/datadog_lambda/patch.py
+++ b/datadog_lambda/patch.py
@@ -9,6 +9,7 @@ import sys
 import logging
 
 from wrapt import wrap_function_wrapper as wrap
+from wrapt.importer import when_imported
 
 from datadog_lambda.tracing import get_dd_trace_context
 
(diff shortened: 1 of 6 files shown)


datadog_lambda/patch.py
```diff
@@ -66,9 +80,9 @@ def _wrap_requests_request(func, instance, args, kwargs):
     into the outgoing requests.
     """
     context = get_dd_trace_context()
-    if "headers" in kwargs:
+    if "headers" in kwargs and isinstance(kwargs["headers"], dict):
         kwargs["headers"].update(context)
-    elif len(args) >= 5:
+    elif len(args) >= 5 and isinstance(args[4], dict):
```
agocs: goodness

PR #53, opened 2020-03-24
Darcy.rayner/dd trace support

DarcyRaynerDD:
### What does this PR do?

Adds beta dd-trace support to python lambda layer. Two new environment variables are added DD_TRACE_ENABLED (enables dd-trace), and DD_MERGE_XRAY_TRACES, (reparents dd-trace traces to xray trace). Currently waiting for latest version of dd-trace-py with required changes to be available.

### Testing Guidelines

I modified the integration tests to capture dd trace output

### Additional Notes

### Checklist

- [ ] Update README version numbers
- [ ] Update dd-trace dependency


datadog_lambda/wrapper.py
```diff
@@ -105,19 +116,47 @@ def __call__(self, event, context, **kwargs):
 
     def _before(self, event, context):
         try:
+
             set_cold_start()
             submit_invocations_metric(context)
             # Extract Datadog trace context from incoming requests
-            extract_dd_trace_context(event)
+            dd_context = extract_dd_trace_context(event)
+            span_context = None
```
tianchu: Nits: Can we refactor these into a helper function rather than polluting the wrapper code with too much details?

datadog_lambda/tracing.py
```diff
@@ -41,6 +50,42 @@ def _convert_xray_sampling(xray_sampled):
...
+    xray_trace_entity = xray_recorder.get_trace_entity()  # xray (sub)segment
+    return {
+        "trace-id": _convert_xray_trace_id(xray_trace_entity.trace_id),
+        "parent-id": _convert_xray_entity_id(xray_trace_entity.id),
+        "sampling-priority": _convert_xray_sampling(xray_trace_entity.sampled),
+        "source": Source.XRAY,
+    }
+
+
+def _get_dd_trace_native_context():
```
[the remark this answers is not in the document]
DarcyRaynerDD: There is already get_dd_trace_context in this file, which I think is part of our public api. Maybe I could rename this function. to _get_dd_trace_py_context?

datadog_lambda/tracing.py
```diff
@@ -41,6 +50,42 @@ def _convert_xray_sampling(xray_sampled):
...
+    span = tracer.current_span()
+    if not span:
+        return None
+
+    parent_id = span.context.span_id
+    trace_id = span.context.trace_id
+    return {
+        "parent-id": str(parent_id),
+        "trace-id": str(trace_id),
+        "sampling-priority": SamplingPriority.AUTO_KEEP,
```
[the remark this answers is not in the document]
DarcyRaynerDD: Good catch. Looking at the JS code, it will keep the sampling priority from the span.

Jwan622:
What will this give us? What metrics? Does it just show us how often other methods are called successfully, how often other methods error, and invocation time? If we wanted metrics that are more granular such as when conditional branches of our lambda code are executed, are custom metrics still the best way?

DarcyRaynerDD:
@Jwan622, this is support for using datadog's own tracing library, ddtrace, in lambda. Previously we just supported using the X-Ray sdk for tracing, (which we will continue to suport). The main difference is ddtrace is open tracing compliant and has integrations outside of the AWS ecosystem. We also can provide you better sampling controls.

README.md
```diff
@@ -265,6 +267,35 @@ def lambda_handler(event, context):
...
+1. If you are using the pip package `datadog-lambda-python`, upgrade it to at least version `v0.X.0`.
+1. Install (or update to) the latest version of [Datadog forwarder Lambda function](https://docs.datadoghq.com/integrations/amazon_web_services/?tab=allpermissions#set-up-the-datadog-lambda-function). Ensure the trace forwarding layer is attached to the forwarder, e.g., ARN for Python 2.7 `arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Trace-Forwarder-Python27:4`.
+1. Set the environment variable `DD_TRACE_ENABLED` to true on your function.
+1. Instrument your function using `dd-trace`.
+
+```py
+from datadog_lambda.metric import lambda_metric
+from datadog_lambda.wrapper import datadog_lambda_wrapper
+
+from ddtrace import tracer
```
tianchu, line 284: Can we add an empty line here?

tests/Dockerfile
```diff
@@ -8,4 +8,5 @@ WORKDIR /test
 
 # Install datadog-lambda with dev dependencies from local
 COPY . .
-RUN pip install .[dev]
\ No newline at end of file
+RUN pip install .[dev]
+RUN pip install --find-links=https://s3.amazonaws.com/pypi.datadoghq.com/trace-dev/index.html ddtrace==0.35.1.dev5+g1e11b2dd -t .[dev]
```
tianchu, line 12: Should we use ddtrace 0.36.0?

datadog_lambda/wrapper.py
```diff
@@ -81,13 +86,21 @@ def __init__(self, func):
...
+            self.propagator = HTTPPropagator()
 
             # Inject trace correlation ids to logs
             if self.logs_injection:
                 inject_correlation_ids()
 
-            # Patch HTTP clients to propagate Datadog trace context
-            patch_all()
+            if not dd_tracing_enabled:
+                # When using dd_trace_py it will patch all the http clients for us,
```
tianchu, line 101: If using ddtrace, does user still need to explicitly call `patch_all` in their function code? If yes, is there any downside for us to call the ddtrace patch_all for here?

datadog_lambda/wrapper.py
```diff
@@ -105,24 +118,59 @@ def __call__(self, event, context, **kwargs):
...
         try:
+            if self.span:
+                self.span.finish()
             if not self.flush_to_log:
                 lambda_stats.flush(float("inf"))
             logger.debug("datadog_lambda_wrapper _after() done")
         except Exception:
             traceback.print_exc()
 
+    def _create_dd_trace_py_span(self, context, trace_context):
```
tianchu, line 147: Nits, can we move this function to tracing.py and pass in needed `self.*` attributes as parameters, so tracing related code all live in the same place?

tianchu: Or define this logic as a class in tracing.py and mix it into the wrapper class here?

datadog_lambda/wrapper.py
```diff
@@ -105,24 +118,59 @@ def __call__(self, event, context, **kwargs):
...
+
             set_cold_start()
             submit_invocations_metric(context)
             # Extract Datadog trace context from incoming requests
-            extract_dd_trace_context(event)
+            dd_context = extract_dd_trace_context(event)
+
+            self.span = None
+            if dd_tracing_enabled:
+                self.span = self._create_dd_trace_py_span(context, dd_context)
```
tianchu, line 129: Nits, i feel this function is more about "creating a span for the function execution" rather than "creating a dd trace py span".  Perhaps `create_function_execution_span` or something better?

tests/Dockerfile
DarcyRaynerDD: Oh, I forgot to remove this from the test docker file. This line was just temporary until the ddtrace-py release came out

datadog_lambda/wrapper.py
```diff
@@ -105,24 +118,59 @@ def __call__(self, event, context, **kwargs):
...
+        tags = {}
+        if context:
+            tags = {
+                "cold_start": is_cold_start(),
+                "function_arn": context.invoked_function_arn,
+                "request_id": context.aws_request_id,
+                "resource_names": context.function_name,
+            }
+        args = {
+            "service": self.function_name,
```
tianchu, line 165: Let's default service name to `aws.lambda` for future proof. Going to look a little bit odd (starting a new service) if merge_xray_traces is true, but I think that' probably fine, and user should be able to override the service name using a tag to combine the two xray spans and this dd span under one service.

datadog_lambda/wrapper.py
```diff
@@ -105,24 +118,59 @@ def __call__(self, event, context, **kwargs):
...
+            "resource": self.handler_name,
```
tianchu, line 166: To be future proof, let' use `function_name` as resource.

datadog_lambda/wrapper.py
DarcyRaynerDD: Should we do this now, if we don't yet support the service tag override? Also, this won't appear on the trace view yet, since it searches by the service name, not the resource name.

datadog_lambda/wrapper.py
DarcyRaynerDD: I've added the ddtrace.patch_all call here

datadog_lambda/wrapper.py
DarcyRaynerDD: I've moved this logic out into the tracing file

PR #56, opened 2020-04-09
Patching http clients erase existing headers

tianchu:
### What does this PR do?

https://github.com/DataDog/datadog-lambda-layer-python/commit/ebeee64c99073db1f80ede8182cdf3dcfc67ff48 introduced a bug into the http clients patching -- existing headers may potentially be erased when injecting Datadog trace context headers, and only Datadog trace context headers will be sent with the request.

This is because some clients store headers in a dict-like objects, such as [requests.structures.CaseInsensitiveDict](https://github.com/psf/requests/blob/bbc3d43522bcb1363050557655508d66bfebcfcf/requests/structures.py#L15) and [botocore.awsrequest.HeadersDict](https://github.com/boto/botocore/blob/ce4ad89e7212026099f1e029249d274e6e1842bc/botocore/awsrequest.py#L591). These objects won't pass the `isinstance(..., dict)` test.

Instead of testing `isinstance(..., dict)`, we should test `isinstance(..., MutableMapping)`.

### Checklist


tests/integration/snapshots/logs/http-requests_python27.log
```diff
@@ -1,27 +1,21 @@
...
...ics": {"_sampling_priority_v1": 1, "system.pid": XXXX, "_dd.agent_psr": 1.0}, "parent_id": "XXXX", "meta": {"function_arn": "arn:aws:lambda:us-east-1:601427279990:function:integration-tester-dev-http-requests_python27", "resource_names": "integration-tester-dev-http-requests_python27", "cold_start": "false", "_dd.origin": "lambda", "_dd.parent_source": "xray", "request_id": "XXXX"}, "error": 0, "duration": XXXX, "type": "serverless", "span_id": "XXXX"}, {"resource": "requests.request", "name": "requests.request", "service": "aws.lambda", "start": XXXX, "trace_id": "XXXX", "metrics": {"_dd.measured": 1}, "parent_id": "XXXX", "meta": {"http.status_code": "200", "http.url": "https://ip-ranges.datadoghq.com/", "_dd.origin": "lambda", "http.method": "GET"}, "error": 0, "duration": XXXX, "type": "http", "span_id": "XXXX"}, {"resource": "requests.request", "name": "requests.request", "service": "aws.lambda", "start": XXXX, "trace_id": "XXXX", "metrics": {"_dd.measured": 1}, "parent_id": "XXXX", "meta": {"http.status_code": "200", "http.url": "https://ip-ranges.datadoghq.eu/", "_dd.origin": "lambda", "http.method": "GET"}, "error": 0, "duration": XXXX, "type": "http", "span_id": "XXXX"}]]}
```
DarcyRaynerDD, line 10: Doesn't really make sense to me why these are split all of a sudden.

PR #60, opened 2020-05-14
Fix parent source as being set as Xray, when trace merging disabled

DarcyRaynerDD:
### What does this PR do?

Currently, the layer will set a metadata field on the root dd-trace span called '_dd.parent_source'  to 'xray' even when x-ray merging is disabled. That field is used by the trace forwarder to filter out spans, to reduce duplication between x-ray and dd-trace traces.  This PR fixes that issue, so that when x-ray trace merging is disabled, the field remains unset, and the root span is never filtered out.

### Motivation

### Testing Guidelines

### Additional Notes

### Checklist

- [ ] Member of the Datadog team has run integration tests and updated snapshots if necessary

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -234,7 +234,7 @@ def set_dd_trace_py_root(trace_context, merge_xray_traces):
 
 
 def create_function_execution_span(
-    context, function_name, is_cold_start, trace_context
+    context, function_name, is_cold_start, trace_context, merge_xray_traces
 ):
     tags = {}
     if context:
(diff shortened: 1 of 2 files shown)


datadog_lambda/tracing.py
```diff
@@ -234,7 +234,7 @@ def set_dd_trace_py_root(trace_context, merge_xray_traces):
 
 
 def create_function_execution_span(
-    context, function_name, is_cold_start, trace_context
+    context, function_name, is_cold_start, trace_context, merge_xray_traces
```
agocs, line 237: Don't you need to change the signature where this method is called as well?

DarcyRaynerDD: Whoops, good catch. Looks like I had a file I didn't stage into the commit. Should be updated

PR #61, opened 2020-05-21
Add default lambda handler, for zero-code change instrumentation

DarcyRaynerDD:
### What does this PR do?

Adds a 'handler' file, that can be pointed to by the user without having to change their code. Using this file involves two steps.

Set the handler to ''datadog_lambda.handler.handler"
Set 'DD_LAMBDA_HANDLER' to the actual handler

### Motivation

### Testing Guidelines

### Additional Notes

### Checklist

- [ ] Member of the Datadog team has run integration tests and updated snapshots if necessary

### DIFF
--- a/datadog_lambda/handler.py
+++ b/datadog_lambda/handler.py
@@ -1,0 +1,22 @@
+from __future__ import absolute_import
+from importlib import import_module
+import os
+from datadog_lambda.wrapper import datadog_lambda_wrapper
+
+
+class HandlerError(Exception):
+    pass
+
+
+path = os.environ.get("DATADOG_USER_HANDLER", None)
+if path is None:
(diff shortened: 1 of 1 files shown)


datadog_lambda/handler.py
```diff
@@ -0,0 +1,28 @@
...
+
+import os
+from datadog_lambda.wrapper import datadog_lambda_wrapper
+
+
+class HandlerError(Exception):
+    pass
+
+
+path = os.environ.get("DATADOG_USER_HANDLER", None)
```
tianchu, line 17: Let's be consistent with other environment variable and use prefix `DD_`? Also, how about `DD_LAMBDA_HANDLER`?

PR #66, opened 2020-06-11
Add resource and executedversion to enhanced lambda metrics

jcstorms1:
### What does this PR do?

Adds the resource and executedversion tags to the enhanced lambda metrics
Adds new tests 

### Motivation

These tags are already available for lambda metrics and adding them to enhanced metrics provides consistent tagging.

### Testing Guidelines

Unittest in testing script

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tags.py
```diff
@@ -35,16 +35,27 @@ def parse_lambda_tags_from_arn(arn):
     split_arn = arn.split(":")
 
     # If ARN includes version / alias at the end, drop it
-    if len(split_arn) > 7:
-        split_arn = split_arn[:7]
+    if len(split_arn) <= 7:
```
jcstorms1, line 38: Does the ARN ever not contain the version/alias? This might not be necessary if it always has it.

datadog_lambda/tags.py
```diff
@@ -35,16 +35,27 @@ def parse_lambda_tags_from_arn(arn):
...
+        return [
+            "region:{}".format(region),
+            "account_id:{}".format(account_id),
+            "functionname:{}".format(function_name),
+        ]
+
+    else:
+        _, _, _, region, account_id, _, function_name, alias = split_arn
+
+        resource = function_name + "/" + alias
```
DarcyRaynerDD, line 51: I think this should be `function_name + ":" + alias`, based on how cloudwatch metrics with an alias are added. This is based on the metric aws.lambda.invocations with the tag resource:swf-hello-test:aws-alias on the. demo org. Also should resource be added in both cases, just in the first case without the alias. That would make it consistent with cloudwatch

tianchu: It's documented here https://docs.aws.amazon.com/lambda/latest/dg/monitoring-metrics.html. `resource` should be `functionname` (when function is not invoked by version or alias), or `functionname:version` (when invoked by version) or `functionname:alias` (when invoked by alias). The `executedversion` should only be present when the function is invoked by alias (i.e., when resource is `functionname:alias`), so it tells users which version is behind that alias.

That said, I believe ARN won't give you everything you need, you probably also need the `function_version` field from the `context` object https://docs.aws.amazon.com/lambda/latest/dg/python-context.html.

datadog_lambda/tags.py
```diff
@@ -63,6 +74,7 @@ def get_enhanced_metrics_tags(lambda_context):
     return parse_lambda_tags_from_arn(lambda_context.invoked_function_arn) + [
         get_cold_start_tag(),
         "memorysize:{}".format(lambda_context.memory_limit_in_mb),
+        "executedversion:{}".format(lambda_context.function_version),
```
tianchu, line 77: See my other comment, to be consistent with CloudWatch metrics, can you make `executedversion` only present when the function is invoked by alias?

PR #77, opened 2020-08-25
Get list of regions from AWS

hghotra:
### What does this PR do?

Get list of regions from AWS rather than having a static list.

### Testing Guidelines

Tested manually.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/publish_layers.sh
+++ b/scripts/publish_layers.sh
@@ -16,7 +16,7 @@ trap "pkill -P $$; exit 1;" INT
 PYTHON_VERSIONS_FOR_AWS_CLI=("python2.7" "python3.6" "python3.7" "python3.8")
 LAYER_PATHS=(".layers/datadog_lambda_py2.7.zip" ".layers/datadog_lambda_py3.6.zip" ".layers/datadog_lambda_py3.7.zip" ".layers/datadog_lambda_py3.8.zip")
 AVAILABLE_LAYER_NAMES=("Datadog-Python27" "Datadog-Python36" "Datadog-Python37" "Datadog-Python38")
-AVAILABLE_REGIONS=(us-east-2 us-east-1 us-west-1 us-west-2 ap-east-1 ap-south-1 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-northeast-1 ca-central-1 eu-north-1 eu-central-1 eu-west-1 eu-west-2 eu-west-3 sa-east-1)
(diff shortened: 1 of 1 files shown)


scripts/publish_layers.sh
```diff
@@ -85,7 +85,7 @@ publish_layer() {
     echo "Published layer for region $region, python version $aws_version_key, layer_name $layer_name, layer_version $version_nbr"
 }
 
-BATCH_SIZE=60
+BATCH_SIZE=1
```
agocs, line 88: Is the goal here to make publishing serial by default, or to remove parallel publishing altogether? If it's the latter, I don't see a reason why we'd want to keep this extra parallel publishing code. If it's the former, do we document a way to force parallel publishing?

scripts/publish_layers.sh
```diff
@@ -16,7 +16,7 @@ trap "pkill -P $$; exit 1;" INT
 PYTHON_VERSIONS_FOR_AWS_CLI=("python2.7" "python3.6" "python3.7" "python3.8")
 LAYER_PATHS=(".layers/datadog_lambda_py2.7.zip" ".layers/datadog_lambda_py3.6.zip" ".layers/datadog_lambda_py3.7.zip" ".layers/datadog_lambda_py3.8.zip")
 AVAILABLE_LAYER_NAMES=("Datadog-Python27" "Datadog-Python36" "Datadog-Python37" "Datadog-Python38")
-AVAILABLE_REGIONS=(us-east-2 us-east-1 us-west-1 us-west-2 ap-east-1 ap-south-1 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-northeast-1 ca-central-1 eu-north-1 eu-central-1 eu-west-1 eu-west-2 eu-west-3 sa-east-1)
+AVAILABLE_REGIONS=$(aws ec2 describe-regions | jq -r '.[] | .[] | .RegionName')
```
agocs, line 19: same, maybe wanna avoid this if $1 is given

scripts/publish_layers.sh
hghotra: The goal is to make publishing serial by default to ensure slow network connections don't cause timeout issues. If someone feels like their internet speed can handle parallel uploads then they can increase the batch size to force parallel publishing. I can document this in our internet docs as note when running `scripts/publish_prod.sh`.

PR #101, opened 2020-12-08
Extract span tags from triggering event

DylanLovesCoffee:
### What does this PR do?

Parses the event invoking the function for tags to be added to the span. Currently only adds the following services to be detected: API Gateway, ALB, SQS, SNS, DynamoDB, Kinesis, Cloudfront, CW Logs, CW Events. API Gateway and ALB trigger events will also add the HTTP url, method, path, referer and status code response if available.

For the case of X-Ray <> Datadog hybrid tracing, we create an additional dummy Datadog subsegment for the trigger tags.

### Motivation

SLS-790 - To give information on what triggered the function execution in one's traces.

### Testing Guidelines

New unit test and edited an existing one. Updated snapshots.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/trigger.py
```diff
@@ -0,0 +1,182 @@
...
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2019 Datadog, Inc.
+
+import base64
+import gzip
+import json
+from io import BytesIO, BufferedReader
+
+
+EVENT_SOURCES = {
```
tianchu, line 12: This feels like a mapping rather than a list of event sources. In fact, the mapping has a very clear pattern as well, so I would simply do a `replace("aws:", "")` in `read_event_source` instead.

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,182 @@
...
+}
+
+
+def get_first_record(event):
+    records = event.get("Records")
+    if records and len(records) > 0:
+        return records[0]
+
+
+def read_event_source(event):
```
tianchu, line 27: Can you make the function name more explicit/specific? I don't see a good difference between `read_event_source` and `get_event_source`, and `get_event_source` calling `read_event_source` doesn't really tell a good story on what is the relationship here.

tianchu: Actually, do you even need this function at all? It might be better to just put these code back into get_event_source?

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,182 @@
...
+    event_source = event.get("eventSource") or event.get("EventSource")
+    return EVENT_SOURCES.get(event_source)
+
+
+def get_event_source(event):
```
tianchu, line 32: Can you add a docstring summarizing all the event sources it might return?

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,182 @@
...
+    if request_context and request_context.get("stage"):
+        event_source = "api-gateway"
+
+    if event.get("awslogs"):
+        event_source = "cloudwatch-logs"
+
+    event_detail = event.get("detail")
+    cw_event_categories = event_detail and event_detail.get("EventCategories")
+    if event.get("source") == "aws.events" or cw_event_categories:
+        event_source = "cloudwatch-events"
```
tianchu, line 48: AWS made this very confusing, just flagging in case this matters https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/WhatIsCloudWatchEvents.html

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,182 @@
...
+        if event_source_arn:
+            trigger_tags["trigger.event_source_arn"] = event_source_arn
+
+    if event_source == "api-gateway":
+        trigger_tags.update(get_apigateway_http_tags(event))
+
+    return trigger_tags
+
+
+def set_apigateway_status_code_tag(span, response):
```
tianchu, line 173: I think the status code might work for ALB as well, either way, I don't think it's necessary to call out apigateway, it might be better to be named as `set_http_status_code_tag`.

tianchu: BTW, did you look into ALB triggered Lambda? https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html

DylanLovesCoffee: Ah, really good point. ALB/ELB totally skipped my mind. It's not in our P1 list but it'll be good to have anyways.

datadog_lambda/wrapper.py
```diff
@@ -138,14 +148,16 @@ def _before(self, event, context):
...
-    def _after(self, event, context):
+    def _after(self, event, context, response):
         try:
             if not self.flush_to_log or should_use_extension:
                 lambda_stats.flush(float("inf"))
             if should_use_extension:
                 flush_extension()
 
             if self.span:
+                if response:
```
tianchu, line 159: nits: this `if` seems a little bit redundant and an over protection.

datadog_lambda/wrapper.py
```diff
@@ -103,33 +106,40 @@ def __init__(self, func):
...
             set_cold_start()
             submit_invocations_metric(context)
-            # Extract Datadog trace context from incoming requests
-            dd_context = extract_dd_trace_context(event)
+            # Extract trigger tags
+            trigger_tags = extract_trigger_tags(event, context)
+            # Extract Datadog trace context and source from incoming requests
+            dd_context, trace_context_source = extract_dd_trace_context(event)
+            if trace_context_source == TraceContextSource.EVENT:
+                create_dd_metadata_subsegment(event, dd_context, trigger_tags)
```
tianchu, line 131: Actually I think the current code is problematic -- we used to only subsegment to pass the datadog trace context when the incoming http event carries a Datadog trace context, but I think you actually need to **always** create the subsegment whenever you can extract trigger tags from the event?

datadog_lambda/tracing.py
```diff
@@ -89,18 +88,37 @@ def _context_obj_to_headers(obj):
     }
 
 
+def create_dd_metadata_subsegment(event, dd_context, trigger_tags):
```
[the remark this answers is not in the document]
DylanLovesCoffee: 🤦

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,191 @@
...
+        return event_record.get("s3")["bucket"]["arn"]
+
+    # e.g. arn:aws:sns:us-east-1:123456789012:sns-lambda
+    if source == "sns":
+        return event_record.get("Sns")["TopicArn"]
+
+    # e.g. arn:aws:cloudfront::123456789012:distribution/ABC123XYZ
+    if source == "cloudfront":
+        distribution_id = event_record.get("cf")["config"]["distributionId"]
+        return "arn:aws:cloudfront::{}:distribution/{}".format(
```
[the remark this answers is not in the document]
DylanLovesCoffee: Ah really good catch

datadog_lambda/wrapper.py
DylanLovesCoffee: Ah that's right. My thinking was that we only needed to add trigger tags when hybrid tracing was in use. I updated this with a function to add generic tags into the `root_span_metadata` field.

datadog_lambda/trigger.py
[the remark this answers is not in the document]
DylanLovesCoffee: Updated this with a function to determine if the ARN should be standard, govcloud or China syntax for the ARNs we stitch together.

datadog_lambda/tracing.py
```diff
@@ -89,18 +88,45 @@ def _context_obj_to_headers(obj):
     }
 
 
+def create_dd_metadata_subsegment(dd_context):
```
tianchu, line 91: nits: create_dd_metadata_subsegment -> create_dd_trace_context_metadata_subsegment

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,205 @@
...
+    return "aws"
+
+
+def get_first_record(event):
+    records = event.get("Records")
+    if records and len(records) > 0:
+        return records[0]
+
+
+def get_event_source(event):
```
tianchu, line 38: nits: parse_event_source?

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,205 @@
...
+    "aws:s3",
+    "aws:sns",
+    "aws:sqs",
+]
+
+GOVCLOUD_REGIONS = ["us-gov-east-1", "us-gov-west-1"]
+CHINA_REGIONS = ["cn-north-1", "cn-northwest-1"]
+
+
+def get_arn_region_identifier(region):
```
tianchu, line 24: nits: the formal name is partition, `get_aws_partition_by_region` https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,205 @@
...
+
+EVENT_SOURCES = [
+    "aws:dynamodb",
+    "aws:kinesis",
+    "aws:s3",
+    "aws:sns",
+    "aws:sqs",
+]
+
+GOVCLOUD_REGIONS = ["us-gov-east-1", "us-gov-west-1"]
```
tianchu, line 20: Any chance you can get the partition directly from the `event` or `context` object? Otherwise it's guaranteed that you would have to update these lists in the future. The worst case, can you use the region prefix, say `us-gov-` or `cn-` to determine partition?

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,205 @@
...
+            "eventSourceArn"
+        )
+
+    if event_source_arn is None:
+        event_source_arn = parse_event_source_arn(source, event, context)
+
+    return event_source_arn
+
+
+def get_http_tags(event):
```
tianchu, line 145: nits: `extract_http_tags`? One way to determine a good function name is to reduce it from the docstring, in this case `Extracts HTTP facet tags from the triggering event`.

datadog_lambda/wrapper.py
```diff
@@ -103,33 +107,42 @@ def __init__(self, func):
...
+            self._after(event, context, response)
 
     def _before(self, event, context):
         try:
 
             set_cold_start()
             submit_invocations_metric(context)
-            # Extract Datadog trace context from incoming requests
-            dd_context = extract_dd_trace_context(event)
+            # Extract trigger tags and add them to a subsegment if found
```
tianchu, line 127: nits: I would shift some of the logic to an intermediary function, say `create_dd_subsegments`, to keep this entry function better manifest the high-level logic and not overloaded/overwhelmed by details. 

```python
trigger_tags = extract_trigger_tags(event, context)
dd_context, trace_context_source = extract_dd_trace_context(event)
create_dd_subsegments(trigger_tags, dd_context, trace_context_source)
```

BTW, perhaps by doing so, you could even avoid creating **two** subsegments (is it necessary?)? ie., `create_dd_subsegment` creates one subsegment, and add the trace context or root span tags based on the need?

datadog_lambda/wrapper.py
```diff
@@ -103,33 +107,42 @@ def __init__(self, func):
...
+            trigger_tags = extract_trigger_tags(event, context)
+            if trigger_tags:
+                create_dd_root_span_metadata_subsegment(trigger_tags)
```
tianchu, line 130: BTW, just a reminder, i recall you would have to duplicate the root span tags to the `aws.lambda.function` span as well, right?

datadog_lambda/trigger.py
DylanLovesCoffee: TIL!

datadog_lambda/wrapper.py
DylanLovesCoffee: Correct 👍

datadog_lambda/trigger.py
DylanLovesCoffee: Yeah that's a really good point. The `event` object may not have the partition but we parse out the `region` from the Lambda `context` so it will probably work best to check the region starts with either of the partitions.

datadog_lambda/tracing.py
```diff
@@ -88,33 +88,44 @@ def _context_obj_to_headers(obj):
...
+def create_dd_subsegment(dd_context, trace_context_source, lambda_function_tags):
+    """
+    Create a Datadog subsegment to pass the Datadog trace context or tags into
+    its metadata field
+    """
+    if not dd_context and not lambda_function_tags:
+        return
+    xray_recorder.begin_subsegment(XraySubsegment.NAME)
+
+    create_dd_root_span_metadata_subsegment(lambda_function_tags)
```
tianchu, line 125: nits: I would probably rename it to `add_dd_root_span_metadata` to reflect what it does now.

datadog_lambda/tracing.py
```diff
@@ -88,33 +88,44 @@ def _context_obj_to_headers(obj):
...
+    if trace_context_source == TraceContextSource.EVENT:
+        create_dd_trace_context_metadata_subsegment(dd_context)
```
tianchu, line 127: nits: same here, `add_dd_trace_context_metadata` seems more accurate.

datadog_lambda/trigger.py
```diff
@@ -0,0 +1,211 @@
...
+    as a tag to the function execution span.
+    """
+    is_http_trigger = span and span.get_tag("trigger.event_source") == (
+        "api-gateway" or "application-load-balancer"
+    )
+    if not is_http_trigger:
+        return
+
+    status_code = "200"
+    if response is None:
```
tianchu, line 205: What actually happens when a Lambda function returns `None` as response to API Gateway? Does that actually cause API Gateway to return 502?

tianchu: I think you assumed that `response` is `None` only if the function throws an exception [here](https://github.com/DataDog/datadog-lambda-python/pull/101/files#diff-5efc779a7e3c73ebd812eb62ff5dce68db1e0ed9ef9765bb076638733e8c5ae6R110)? But what if the customer's handler function somehow does return `None` as response?

DylanLovesCoffee: Yup, good thought -- it seems that if the Lambda handler returns `None`, API Gateway will always return a 502 code. I think that relates to the "incompatible output" here: https://docs.aws.amazon.com/apigateway/api-reference/handling-errors/

tianchu (approved):
👍

datadog_lambda/wrapper.py
```diff
@@ -138,15 +148,18 @@ def _before(self, event, context):
...
             if not self.flush_to_log or should_use_extension:
                 lambda_stats.flush(float("inf"))
             if should_use_extension:
                 flush_extension()
 
             if self.span:
+                set_http_status_code_tag(self.span, response)
                 self.span.finish()
+                # Close the DD subsegment here and send it
+            close_dd_subsegment()
```
DylanLovesCoffee, line 162: @tianchu I updated this to end the DD subsegment in the `_after` block so that we can pass the `status_code` tag to the existing subsegment. Is that the right way to go about it?

datadog_lambda/tracing.py
```diff
@@ -89,18 +88,73 @@ def _context_obj_to_headers(obj):
...
+def end_dd_dummy_metadata_subsegment(dd_dummy_subsegment):
+    """
+    End the Datadog dummy metadata subsegment if it is the current subsegment
+    """
+    if dd_dummy_subsegment is None:
+        return
+    # Ensure we are ending the Datadog dummy subsegment
+    subsegment = xray_recorder.current_subsegment()
+    if subsegment is dd_dummy_subsegment:
+        xray_recorder.end_subsegment()
```
DylanLovesCoffee, line 146: Added a check for the current subsegment since I wasn't able to end the Datadog dummy subsegment directly.

datadog_lambda/wrapper.py
```diff
@@ -102,34 +105,42 @@ def __init__(self, func):
 
     def __call__(self, event, context, **kwargs):
         """Executes when the wrapped function gets called"""
-        self._before(event, context)
+        trigger_tags = extract_trigger_tags(event, context)
```
tianchu, line 108: Nits -- I think it's better to put `trigger_tags` and `response` onto `self` (similar to `self.span`), so you don't have to pass them to `_before` or `_after`. TBH, we could have apply the same pattern to `event` and `context`, but they seem to be important enough to standout.

datadog_lambda/wrapper.py
```diff
@@ -138,14 +149,24 @@ def _before(self, event, context):
         except Exception:
             traceback.print_exc()
 
-    def _after(self, event, context):
+    def _after(self, event, context, response, trigger_tags):
         try:
+            status_code = extract_http_status_code_tag(trigger_tags, response)
+            if status_code:
+                trigger_tags["http.status_code"] = status_code
+            # Create a Datadog subsegment with trigger tags
```
tianchu, line 157: I don't think the future reader will be able to understand what really does or why it's needed from this comment.

datadog_lambda/tracing.py
```diff
@@ -90,6 +89,27 @@ def _context_obj_to_headers(obj):
...
 
+def create_dd_dummy_metadata_subsegment(
+    dd_context, trace_context_source, subsegment_metadata_key
+):
+    """
+    Create a Datadog subsegment to pass the Datadog trace context or Lambda function
+    tags into its metadata field, so the X-Ray trace can be converted to a Datadog
+    trace in the Datadog backend with the correct context.
+    """
+    if not dd_context or (
```
tianchu, line 100: With the refactoring, `create_dd_dummy_metadata_subsegment` becomes a more general helper function, I don't think `dd_context` is still the right argument name. Perhaps it should be `subsegment_metadata_value`?

datadog_lambda/tracing.py
```diff
@@ -90,6 +89,27 @@ def _context_obj_to_headers(obj):
...
+        trace_context_source and trace_context_source != TraceContextSource.EVENT
```
tianchu, line 101: I not sure this check on `trace_context_source` belongs to this general helper function, e.g., is it needed for both use cases? 

I think you should do the check outside before calling this function.

datadog_lambda/wrapper.py
```diff
@@ -102,34 +105,42 @@ def __init__(self, func):
...
         try:
 
             set_cold_start()
             submit_invocations_metric(context)
-            # Extract Datadog trace context from incoming requests
-            dd_context = extract_dd_trace_context(event, context)
+            # Extract Datadog trace context and source from incoming requests
+            dd_context, trace_context_source = extract_dd_trace_context(event, context)
+            # Create a Datadog X-Ray subsegment with the trace context
+            create_dd_dummy_metadata_subsegment(
```
tianchu, line 130: Related to the previous comment. I think you should call this function `if dd_context and trace_context_source == TraceContextSource.EVENT`?

datadog_lambda/wrapper.py
```diff
@@ -138,14 +149,24 @@ def _before(self, event, context):
...
             traceback.print_exc()
 
-    def _after(self, event, context):
+    def _after(self, event, context, response, trigger_tags):
         try:
+            status_code = extract_http_status_code_tag(trigger_tags, response)
+            if status_code:
+                trigger_tags["http.status_code"] = status_code
+            # Create a Datadog subsegment with trigger tags
+            create_dd_dummy_metadata_subsegment(
```
tianchu, line 158: Related to the comment on `create_dd_dummy_metadata_subsegment`. I think you want to call this function `if trigger_tags:`?

datadog_lambda/wrapper.py
DylanLovesCoffee: Much cleaner now 👍

PR #112, opened 2021-01-15
[POC] Add `DD_TRACE_EXTRACTOR` env var to customize Trace extraction customization

Czechh:
### What does this PR do?
Analogous to https://github.com/DataDog/datadog-lambda-js/pull/151

Provides a way to customize how to extract the trace from the event or context. This will allow us to quickly support new tools and technologies with custom code.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/wrapper.py
```diff
@@ -91,6 +93,16 @@ def __init__(self, func):
                 os.environ.get("DD_MERGE_XRAY_TRACES", "false").lower() == "true"
             )
             self.function_name = os.environ.get("AWS_LAMBDA_FUNCTION_NAME", "function")
+            self.extractor_env = os.environ.get("DD_TRACE_EXTRACTOR", None)
+            self.trace_extractor = None
+
+            if self.extractor_env:
```
DarcyRaynerDD, line 99: If someone isn't using the redirected handler, could this cause a cyclical dependency. Eg. user does this:
```python
from datadog_lambda.wrapper import datadog_lambda_wrapper

@datadog_lambda_wrapper
def my_lambda_handler(event, context):
   pass

def my_lambda_extractor(event, context):
   //... some extractor logic here
```

In this case, the import chain would be my_function.py -> datadog_lambda.wrapper -> my_function.py -> etc etc .

DarcyRaynerDD: In node we addressed this by only supporting the env var for the extractor in the redirected handler. When you aren't using the redirected handler, it was possible to pass the extractor in as a config variable to the wrapper.

tianchu: I think I can probably delay the import to when the extractor needs to be invoked. But I'd rather simply instruct users to put the extractor in a separate file (from the one where the handler is defined). I think that's probably good enough for a poc and given the fact that most of applications should be using the redirected handler now.

DarcyRaynerDD (approved):

PR #120, opened 2021-03-29
Disable flushing metrics in thread

tianchu:
### What does this PR do?

Disable `flushing_in_thread` of `threadstats` by default. Previously, for synchronous metric submission (without piggybacking on the forwarder and logs), the `threadstats` is configured to flush the metrics both at the end of the invocation (blocking) and every 10s in a background thread (non-blocking). This PR disables the periodic background flushing.

This change is a soft breaking change, though it shouldn't result in any data loss, it does delay synchronous metric submission for long-running Lambda functions (duration > 10s). In the worst case, metrics submitted at the beginning of an invocation could be delayed for showing up in Datadog for 15 min (the max Lambda duration allowed).  

For affected functions, the behavior could be re-enabled by setting env var `DD_FLUSH_IN_THREAD` to `true`. However, given the problematic nature of this flushing behavior, I would recommend submitting metrics through the new [Datadog Lambda extension](https://docs.datadoghq.com/serverless/datadog_lambda_library/extension/) instead.

[Datadog Forwarder](https://github.com/DataDog/datadog-serverless-functions/tree/master/aws/logs_monitoring) relies on the `datadog-lambda` to flush metrics, so the `datadog-lambda` version needs to be bumped to get the issue fixed for the Datadog Forwarder https://github.com/DataDog/datadog-serverless-functions/pull/436

Also fixed the broken and flaky integration tests.

### Motivation

A bug has been recently identified (see sample logs below). If there happen to be a background flushing in progress, the end-of-invocation flushing will be skipped (because there is already an in-flight flushing).  However, if the background flushing doesn't finish before the AWS freezing the execution environment, it will resume at the beginning of the next invocation. There are two problems with it. First, if AWS decides to destroy the current execution environment, this request will never complete. Second, if the next invocation only happens a few min later (e.g., a cronjob), the request will fail due to connection timeout.

In short, the background thread flushing isn't suitable for Lambda and should be disabled.

```
# current invocation
Flushing metrics in thread
Flush #8 sending 640 distributions
A flush is already in progress. Skipping this one.
datadog_lambda_wrapper _after() done
REPORT RequestId: d0deabde-936b-467c-aba6-7d4cebdf25c0	Duration: 843.65 ms	Billed Duration: 844 ms	Memory Size: 1024 MB	Max Memory Used: 146 MB	
END RequestId: d0deabde-936b-467c-aba6-7d4cebdf25c0

# next invocation (5 mins later)
START RequestId: faa32e6f-8cd4-4384-aac1-2a5698344e54 Version: $LATEST
Could not request POST https://api.datadoghq.com/api/v1/distribution_points: ('Connection aborted.', timeout('The write operation timed out')). Please check the network connection or try again later.
``` 

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change

### Check all that apply

- [x] This PR contains breaking changes that are documented in the description
- [x] This PR impacts documentation, and it has been updated (or a ticket has been logged)


README.md
```diff
@@ -87,6 +87,10 @@ Initialize the Datadog tracer when set to `true`. Defaults to `false`.
 
 Set to `true` to merge the X-Ray trace and the Datadog trace, when using both the X-Ray and Datadog tracing. Defaults to `false`.
 
+### DD_FLUSH_IN_THREAD
+
+Flush metrics periodically in a background thread. See details in this [PR](https://github.com/DataDog/datadog-lambda-python/pull/120). Defaults to `false`.
+
```
nhinsch, line 93: I wouldn't include a link to a PR in the README, I would just explain directly in the README.

With that said, why do we have this as an environment variable? In the PR description, you wrote "In short, the background thread flushing isn't suitable for Lambda and should be disabled." If it's not suitable/should be disabled, I think exposing an option to enable it might lead to confusion with little or no benefit for customers. Maybe we should just force it to be off.

If we really want to preserve the ability to turn this environment variable on for some edge cases, I would recommend that we do one or both of the following:
- Rename the variable to `DD_LEGACY_METRICS_FLUSHING_BEHAVIOR` to discourage people from using it.
- Include the variable in the code and add a comment, but don't include it in the README to avoid confusing new users.

nhinsch (approved):
Looks good, I would just suggest we de-emphasize the setting more, either be removing it completely, renaming it, and/or hiding it, to avoid confusion, since it is discouraged/potentially broken to flush metrics in the background.

Issue #104, opened 2020-12-16
Pinned setuptools version (setuptools==42.0.2) is too restrictive


zlalvani:
## Expected Behavior
Dependency versions are loose enough to be resolved in most cases

## Actual Behavior
Pinned dependency on `setuptools` is old and potentially too specific

## Steps to Reproduce the Problem

  1. `pip install -r requirements.txt` in a requirements file with both datadog-lambda and another dependency with a conflicting `setuptools` version. In this case, `wiremock`. 

## Specifications

  - Datadog Lambda Layer version: 2.26.0
  - Python version: 3.8.6
  
## Stacktrace
  
  ```
ERROR: Cannot install -r requirements-dev.txt (line 13) and -r requirements.txt (line 14) because these package versions have conflicting dependencies.

The conflict is caused by:
    datadog-lambda 2.26.0 depends on setuptools==42.0.2
    wiremock 2.1.3 depends on setuptools>=45.2.0  
```

**Comment by ryboe** [comment]
Yes, we're having issues with Ye Olde Setuptools too.

```
Collecting setuptools==42.0.2
  Downloading setuptools-42.0.2-py2.py3-none-any.whl (583 kB)
ERROR: Hashes are required in --require-hashes mode, but they are missing from some requirements. Here is a list of those requirements along with the hashes their downloaded archives actually had. Add lines like these to your requirements files to prevent tampering. (If you did not enable --require-hashes manually, note that it turns on automatically when any package has a hash.)
    setuptools==42.0.2 --hash=sha256:c8abd0f3574bc23afd2f6fd2c415ba7d9e097c8a99b845473b0d957ba1e2dac6
```

sfirrin:
Hi @zlalvani and @ryboe, thanks for bringing this to our attention. As of https://github.com/DataDog/datadog-lambda-python/releases/tag/v32 we're now specifying the current latest version of setuptools or greater. If you're still running into issues please reopen and let us know.

PR #126, opened 2021-03-31
Use latest setuptools version supported for each Python version

sfirrin:
### What does this PR do?
Upgrades our pinned setuptools to the latest supported for each Python version.

setuptools dropped support for Python 2 in version 45 ([link](https://setuptools.readthedocs.io/en/latest/python%202%20sunset.html)), so we need to specify an older version (44.1.1) for Python 2. This was done using the python_version environment marker documented [here](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies).

### Motivation
https://github.com/DataDog/datadog-lambda-python/issues/104

### Testing Guidelines
Working as expected in the integration tests where we can see the older version is used for Python 2 and the newer version is used for Python 3. 

### Additional Notes
-

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


setup.py
```diff
@@ -32,7 +32,10 @@
         "datadog==0.39.0",
         "ddtrace==0.45.0",
         "wrapt==1.11.2",
-        "setuptools==42.0.2",
+        # If building for Python 3, use the latest version of setuptools
+        "setuptools==54.2.0; python_version >= '3.0'",
```
DarcyRaynerDD, line 36: Should we loosen the version requirement so it's a greater than? I think part of the issue was we lock to a particular version, which the user might not be on.

sfirrin: That sounds good to me, I'm not sure how we'd choose which version though.

I think this would fix this particular customer's issue as is, since the conflict they were getting was
```
The conflict is caused by:
  datadog-lambda 2.26.0 depends on setuptools==42.0.2
  wiremock 2.1.3 depends on setuptools>=45.2.0
```

Which should be fixed by us pinning anything >=45.2.0.

PR #137, opened 2021-04-28
Add the matrix strategy to workflow to make integration tests faster

maxday:
### What does this PR do?

This PR makes the integration testing step faster by using the matrix strategy, it will run tests in parallel.

Current average workflow time for the integration tests step : around 10 minutes
With this PR, the new time is about 4minutes

### Motivation

Increase productivity by reducing waiting time

### Testing Guidelines

Codebase was not change but a refactoring for integration testing has been made. This should not break the integration tests and should be fully backward compatible

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,9 @@ jobs:
   
   integration-test:
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        runtime-param: [27, 36, 37, 38]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -93,4 +95,5 @@ jobs:
           DD_API_KEY: ${{ secrets.DD_API_KEY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          RUNTIME_PARAM: ${{ matrix.runtime-param }}
         run: ./scripts/run_integration_tests.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        runtime-param: [27, 36, 37, 38]
+        runtime-param: [2.7, 3.6, 3.7, 3.8]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


scripts/run_integration_tests.sh
```diff
@@ -52,32 +72,48 @@ input_event_files=$(ls ./input_events)
 # Sort event files by name so that snapshots stay consistent
 input_event_files=($(for file_name in ${input_event_files[@]}; do echo $file_name; done | sort))
 
-# Generate a random 8-character ID to avoid collisions with other runs
-run_id=$(xxd -l 4 -c 4 -p < /dev/random)
-
-# Always remove the stack before exiting, no matter what
+# Always remove the stacks before exiting, no matter what
```
nhinsch, line 75: Nit: I see you changed stack -> stacks here, was that intentional? Isn't this still going to deploy one stack at a time?

nhinsch: I guess it should now be `stack(s)`

.github/workflows/build.yml
```diff
@@ -62,7 +62,9 @@ jobs:
   
   integration-test:
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        runtime-param: [27, 36, 37, 38]
```
nhinsch, line 67: Nit: Can we use strings with decimals, i.e. "2.7" and "3.6", for these params? I always do a double take when I read "Python 37". Also, that will be more consistent with our unit test task. If it's not possible then no big deal.

nhinsch (approved):
Looks great! 🚀 Just a couple of tiny nits.

scripts/run_integration_tests.sh
maxday: ✅

.github/workflows/build.yml
maxday: ✅

scripts/run_integration_tests.sh
```diff
@@ -38,9 +38,10 @@ PARAMETERS_SETS=("python27" "python36" "python37" "python38")
 if [ -z "$RUNTIME_PARAM" ]; then
     echo "Python version not specified, running for all python versions."
 else
-    echo "Python version is specified: $RUNTIME_PARAM"
-    PARAMETERS_SETS=(python${RUNTIME_PARAM})
-    BUILD_LAYER_VERSION=python$RUNTIME_PARAM[1]
+    RUNTIME_PARAM_NO_DOT=$(echo $RUNTIME_PARAM | sed 's/\.//')
+    echo "Python version is specified: $RUNTIME_PARAM_NO_DOT"
```
nhinsch, line 42: Nit: I think we would probably want to log the version number with the dot here, right? `Python version is specified: 3.7` rather than `Python version is specified: 37`

maxday: ✅

PR #145, opened 2021-06-09
Support decrypting API keys encrypted with an encryption context

nhinsch:
### What does this PR do?

When decrypting the API key using KMS, try first with the encryption context added by the Lambda console UI, then try without.

Bonus: Updates `black` (the code formatter) to v21.

### Motivation

The Lambda console UI changed the way it encrypts environment variables. The current behavior is to encrypt environment variables using the function name as an encryption context. Previously, the behavior was to encrypt environment variables without an encryption context. The Lambda Extension currently only supports API keys encrypted using the previous behavior. We need to support both, as supplying the incorrect encryption context will cause decryption to fail.

### Testing Guidelines

I added unit test coverage for this function and tested manually.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -240,13 +240,15 @@ def decrypt_kms_api_key(kms_client, ciphertext):
             EncryptionContext={
                 KMS_ENCRYPTION_CONTEXT_KEY: function_name,
             },
-        )["Plaintext"].decode('utf-8')
+        )["Plaintext"].decode("utf-8")
     except ClientError:
         logger.debug(
             "Failed to decrypt ciphertext with encryption context, retrying without"
         )
         # Try without encryption context
-        plaintext = kms_client.decrypt(CiphertextBlob=decoded_bytes)["Plaintext"].decode('utf-8')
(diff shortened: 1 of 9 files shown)


datadog_lambda/metric.py
```diff
@@ -213,12 +214,50 @@ def submit_errors_metric(lambda_context):
...
+    """
+    # Try with encryption context
+    function_name = os.environ.get("AWS_LAMBDA_FUNCTION_NAME")
+    try:
+        plaintext = kms_client.decrypt(
+            CiphertextBlob=decoded_bytes,
+            EncryptionContext={
+                KMS_ENCRYPTION_CONTEXT_KEY: function_name,
+            },
+        )["Plaintext"]
```
tianchu, line 243: We don't need something like this? https://github.com/DataDog/datadog-serverless-functions/blob/f2770b49c29d265686f33d70fa9b1f48eca32023/aws/logs_monitoring/settings.py#L205 I thought it's needed in python3?

nhinsch: Good catch!

DarcyRaynerDD (approved):
LGTM

PR #146, opened 2021-06-16
Check for the new direct invocation trace format and use that

agocs:
### What does this PR do?

Updates datadog-lambda-python so it can extract trace contexts stored in `context.client_context.custom` using the new propagation format, described in this doc: https://docs.google.com/document/d/1pruSeesAr2bAVMtz_rUlsiBZKFoFWdB4nSq5Vuym2l8/edit#heading=h.ptr8uhqw5ht

### Motivation

Storing a map containing trace context in `context.client_context.custom._datadog` causes compatibility issues with Go and Java lambda functions. C.f. https://github.com/aws/aws-lambda-java-libs/issues/212. We're updating the trace propagation format to fix this issue.

### Testing Guidelines

- Added a unit test

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -6,7 +6,7 @@
 import logging
 import os
 import json
-from typing import Optional
+from typing import Optional  # noqa: F401
 
 from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.core.lambda_launcher import LambdaContext
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -6,6 +6,7 @@
 import logging
 import os
 import json
+from typing import Optional
 
 from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.core.lambda_launcher import LambdaContext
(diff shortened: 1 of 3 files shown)


datadog_lambda/tracing.py
```diff
@@ -129,16 +130,27 @@ def extract_context_from_lambda_context(lambda_context):
     dd_trace libraries inject this trace context on synchronous invocations
     """
     client_context = lambda_context.client_context
-
+    trace_id = None  # type: Optional[str]
+    parent_id = None  # type: Optional[str]
+    sampling_priority = None  # type: Optional[str]
```
DarcyRaynerDD, line 135: I think the "str" type is a bit complicated, because it should be 'unicode' in python 2.7, and 'str' in 3+. It seems like the correct type would be 'Text' https://mypy.readthedocs.io/en/stable/cheat_sheet.html#built-in-types

Not sure it's worth introducing type hinting in this PR. I think it would make sense to start at the same time we drop 2.7 support, so we can use the more convenient 3+ syntax

DarcyRaynerDD (approved):
LGTM, left a nit

datadog_lambda/tracing.py
agocs: That makes sense. My instinct is to scatter type hints in as many places as possible, but it's probably better to wait until the 2 -> 3 conversion.

PR #147, opened 2021-06-16
fix hello route hit too early

maxday:
### What does this PR do?

Call the /hello route at the right moment

### Motivation

Consistency in metric tagging with the extension

### Testing Guidelines

With integration testing

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -20,6 +20,7 @@ ENHANCED_METRICS_NAMESPACE_PREFIX = "aws.lambda.enhanced"
 
 logger = logging.getLogger(__name__)
 
+lambda_stats = None
 
 class StatsWriter:
     def distribution(self, metric_name, value, tags=[], timestamp=None):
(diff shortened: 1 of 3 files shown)


nhinsch:
Would you mind explaining the old behavior, why it was wrong, and what the new behavior is? From the diff, I couldn't quite follow how this changed the timing of the `/hello` route.

maxday:
Sure !
Old behaviour : the code responsible to call the `hello` route was located of the root of the metry.py. So as soon as this file gets imported, the code executes. This file is imported in wrapper.py so the code will be executed before the actual invocation. This results in hitting the /hello route before setting the extra tags since tags extraction is done once we received the actual invocation.
New behaviour : the code is within a dedicated function which is triggered after that we have notified the extension of an invocation event, so the extension can set the tag properly and metrics sent from the layer to the extension can be properly enhanced with extra tags.

nhinsch (approved):
LGTM
Is this possible to test in the library, or will we rely on integration tests for this?

maxday:
I think integration tests is a good fit for this one as there are already some unit tests on this part

yury-sannikov:
Well, it seems it's breaking tons of code.

PR #163, opened 2021-08-17
Refactoring to allow lazy loading of datadog.api

maxday:
### What does this PR do?

This PR 
- splits StatsWriter and its subclasses in separate files so only the relevant one is included during cold start
- does not import datadog.api if the extension is used which also speed up the cold start

### Motivation

Cold start improvements

### Testing Guidelines

Unit + Integration tests

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


astuyve (approved):
This looks good to me. Are there other places we're loading boto3 or stats classes, where we could delay it if possible?

maxday:
> This looks good to me. Are there other places we're loading boto3 or stats classes, where we could delay it if possible?

not that I am aware of and I don't see any trace of boto3 loading in the flamegraph (except in aws-xray-sdk but I've opened a PR here https://github.com/aws/aws-xray-sdk-python/pull/312 🤞 )

tianchu:
Should we include something we discussed about https://github.com/DataDog/datadog-lambda-python/pull/147 here? Seems touching similar code path.

datadog_lambda/metric.py
```diff
@@ -24,100 +21,15 @@
...
-            else:
-                logger.debug(
-                    "Flush #%s failed", self.thread_stats.flush_count, exc_info=True
-                )
-
-    def stop(self):
-        self.thread_stats.stop()
-
-
 def init_lambda_stats():
```
tianchu, line 24: As we discussed before, might make sense to call `init_lambda_stats()` in the `__init__` method of the wrapper? So it only creates the writer on cold start?

tianchu: Alternatively you could check `lambda_stats is None` before creating a new writer? But I think it's probably the best that we have both?

datadog_lambda/metric.py
```diff
@@ -257,34 +169,39 @@ def decrypt_kms_api_key(kms_client, ciphertext):
...
-
-logger.debug("Setting DATADOG_API_KEY of length %d", len(api._api_key))
-
-# Set DATADOG_HOST, to send data to a non-default Datadog datacenter
-api._api_host = os.environ.get(
-    "DATADOG_HOST", "https://api." + os.environ.get("DD_SITE", "datadoghq.com")
-)
-logger.debug("Setting DATADOG_HOST to %s", api._api_host)
+# Set API Key only if extension is not here
+if not should_use_extension:
```
tianchu, line 173: Perhaps we should put them into a function? and call it in the wrapper's `__init__` method? I think we can further reduce the scope here, the API key is ONLY needed when `DD_FLUSH_TO_LOG` is `false`.

maxday: I've updated the condition

datadog_lambda/metric.py
maxday: I've reverted the PR #147 as it had no effect on /hello route

PR #170, opened 2021-08-31
Rewrite UPD client when the extension is present to avoid loading datadogpy

maxday:
### What does this PR do?

Do not load `datadogpy` when the extension is detected.
Basically this PR duplicates the Dogstatd from `datadogpy` but in a super light version

### Motivation

Cold start duration improvements

### Testing Guidelines

Unit tests + integration tests

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/statsd_writer.py
+++ b/datadog_lambda/statsd_writer.py
@@ -1,5 +1,5 @@
 from datadog_lambda.stats_writer import StatsWriter
-from datadog import initialize, statsd
+from datadog.dogstatsd import statsd
 
 
 class StatsDWriter(StatsWriter):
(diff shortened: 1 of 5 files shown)


datadog_lambda/statsd_writer.py
```diff
@@ -1,16 +1,12 @@
 from datadog_lambda.stats_writer import StatsWriter
-from datadog import initialize, statsd
+from datadog_lambda.dogstatsd import statsd
```
tianchu, line 2: Can you add a comment in the other place (threadstats_writer?) where datadog is lazy loaded? just so people don't accidentally get import datadog into the critical path in the future?

tianchu (approved):

datadog_lambda/statsd_writer.py
maxday: Good idea! I've updated both files with the comment. In threadstats_writer, there is no need to lazy load `datadog` as this file is lazy loaded itself here : https://github.com/DataDog/datadog-lambda-python/blob/main/datadog_lambda/metric.py#L30

tianchu (approved):

PR #172, opened 2021-09-07
Chris.agocs/inferring api gateway spans poc

agocs:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


agocs:
n.b. @Hesperide the inferred spans are still showing up as the same color as the Lambda spans, and they're the same node on the trace map. I've tried setting the service name to the API Gateway URL (or equivalent), and it doesn't seem to want to take.

Hesperide:
> n.b. @Hesperide the inferred spans are still showing up as the same color as the Lambda spans, and they're the same node on the trace map. I've tried setting the service name to the API Gateway URL (or equivalent), and it doesn't seem to want to take.

How can I help unblock this? Do AJ/Darcy have any ideas?

README.md
```diff
@@ -87,6 +87,16 @@ Initialize the Datadog tracer when set to `true`. Defaults to `false`.
 
 Set to `true` to merge the X-Ray trace and the Datadog trace, when using both the X-Ray and Datadog tracing. Defaults to `false`.
 
+### DD_INFERRED_SPANS (beta)
```
Hesperide, line 90: I don't know if you were planning to update this later, but when we're GA this should default to true.

We should also flag as "experimental" instead of "beta".

tianchu:
> > n.b. @Hesperide the inferred spans are still showing up as the same color as the Lambda spans, and they're the same node on the trace map. I've tried setting the service name to the API Gateway URL (or equivalent), and it doesn't seem to want to take.
> 
> How can I help unblock this? Do AJ/Darcy have any ideas?

This is probably due to the backend tag resolution overriding? I believe the backend applies the service tag value to every span in the same trace payload, because it assumes that spans in the same trace payload come from the same service. One potential solution is to put the inferred trigger span in its own trace payload, perhaps by calling [tracer.write()](https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#ddtrace.Tracer.trace)?

agocs:
> This is probably due to the backend tag resolution overriding? I believe the backend applies the service tag value to every span in the same trace payload, because it assumes that spans in the same trace payload come from the same service. One potential solution is to put the inferred trigger span in its own trace payload, perhaps by calling [tracer.write()](https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#ddtrace.Tracer.trace)?

Hey Tian, I tried explicitly calling `tracer.write()` and it seems to do what we think it does, but the tracer still flushes the whole trace (with both spans) at the end of the execution, to the same effect. 
image

datadog_lambda/wrapper.py
```diff
@@ -180,16 +191,20 @@ def _after(self, event, context):
...
-
             if self.span:
                 if status_code:
                     self.span.set_tag("http.status_code", status_code)
                 self.span.finish()
+            if self.inferred_span:
+                if status_code:
+                    self.inferred_span.set_tag("http.status_code", status_code)
+                self.inferred_span.finish()
             logger.debug("datadog_lambda_wrapper _after() done")
```
tianchu, line 202: This log line is no longer at the right place.

datadog_lambda/tracing.py
```diff
@@ -377,13 +409,125 @@ def set_dd_trace_py_root(trace_context_source, merge_xray_traces):
...
+
+
 def create_function_execution_span(
     context,
     function_name,
     is_cold_start,
     trace_context_source,
     merge_xray_traces,
     trigger_tags,
+    upstream=None,
```
tianchu, line 530: Is "parent" a better name?

datadog_lambda/tracing.py
```diff
@@ -402,6 +546,7 @@ def create_function_execution_span(
             else None,
             "datadog_lambda": datadog_lambda_version,
             "dd_trace": ddtrace_version,
+            "span.name": "aws.lambda",
```
tianchu, line 549: Why do you need this? Doesn't `span = tracer.trace("aws.lambda", **args)` (later in this function) set the span name to `aws.lambda`?

datadog_lambda/tracing.py
```diff
@@ -19,17 +20,48 @@
...
 from datadog_lambda import __version__ as datadog_lambda_version
 
 logger = logging.getLogger(__name__)
 
+
+SPAN_TYPE_TAG = "_dd.span_type"
+SPAN_TYPE_INFERRED = "inferred"
+
+
+class InferredSpanFilter(TraceFilter):
```
tianchu, line 34: I think this class definitely deserve a comprehensive docstring to explain why it's needed and the background.

README.md
```diff
@@ -87,6 +87,17 @@ Initialize the Datadog tracer when set to `true`. Defaults to `false`.
 
 Set to `true` to merge the X-Ray trace and the Datadog trace, when using both the X-Ray and Datadog tracing. Defaults to `false`.
 
+### DD_INFERRED_SPANS (experimental)
+
+Inferred Spans are spans that Datadog can create based on incoming event metadata. 
+Set `DD_INFERRED_SPANS` to `true` to infer spans based on Lambda events.
+Inferring upstream spans is only supported if you are using the [Datadog Lambda Extension](https://docs.datadoghq.com/serverless/libraries_integrations/extension/).
```
DarcyRaynerDD, line 94: Is this actually true? I know we said we wouldn't do any specific work to make it function in the forwarder, but the span will still get generated. I thin originally I said the forwarder might remap the service name, but I don't know if we verified that.

datadog_lambda/tracing.py
```diff
@@ -377,13 +409,125 @@ def set_dd_trace_py_root(trace_context_source, merge_xray_traces):
...
+    except Exception as e:
+        logger.debug(
+            "Unable to infer span. Detected type: {}. Reason: {}", managed_service, e
+        )
+        return None
+    logger.debug("Unable to infer a span: unknown event type")
+    return None
+
+
+def detect_inferrable_span_type(event):
```
tianchu, line 433: Can you reuse (maybe some refactoring required) the same logic from https://github.com/DataDog/datadog-lambda-python/blob/main/datadog_lambda/trigger.py? Want to avoid inconsistency between what's shown by the trace and the invocation.

datadog_lambda/tracing.py
```diff
@@ -377,13 +409,125 @@ def set_dd_trace_py_root(trace_context_source, merge_xray_traces):
...
+    return ManagedService.UNKNOWN
+
+
+def create_inferred_span_from_api_gateway_websocket_event(event, context):
+    domain = event["requestContext"]["domainName"]
+    endpoint = event["requestContext"]["routeKey"]
+    tags = {
+        "operation_name": "aws.apigateway.websocket",
+        "service.name": domain,
+        "url": domain + endpoint,
```
tianchu, line 451: https://github.com/DataDog/datadog-lambda-python/blob/c86be18ec792760e4307653f36cbf5f5c8fd495c/datadog_lambda/trigger.py#L154 sounds like the standard field is `http.url`?

datadog_lambda/tracing.py
DarcyRaynerDD, line 34: I think this warrants a comment explaining why we are using this.

datadog_lambda/tracing.py
DarcyRaynerDD, line 433: Is there a way we can plug this into the existing trigger detection logic. I'm wondering if we are duplicating the functionality somewhat.

DarcyRaynerDD (approved):
Left some nits

datadog_lambda/wrapper.py
```diff
@@ -96,6 +96,11 @@ def __init__(self, func):
             self.extractor_env = os.environ.get("DD_TRACE_EXTRACTOR", None)
             self.trace_extractor = None
             self.span = None
+            self.inferred_span = None
+            self.make_inferred_span = (
+                os.environ.get("DD_INFERRED_SPANS", "false").lower() == "true"
+                and should_use_extension
```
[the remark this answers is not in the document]
agocs: The forwarder adds tags from the tag cache to the trace payload(s) being sent back to the trace intake endpoint. This is an additional variable and an additional thing we would need to fix before the inferred spans show up as their own services. We can always remove this check and add forwarder support later.

README.md
agocs: We did verify it. The forwarder adds tags from the tags cache to the trace payloads it sends back. If we want, we can always fix this and add forwarder support later on.

datadog_lambda/tracing.py
agocs: You and Tian are really on the same brain wave here 😄

README.md
tianchu: hmm, i guess the forwarder applies the service tag on all the trace payloads from the same event object? if so, even the trace payload carrying the inferred span has no lambda function arn, it still assumes it's from the same function with other payloads and apply the tag? If yes, we would need to make some changes in the forwarder, hopefully a small one.

PR #195, opened 2021-11-11
[SLS-1683] Add sync/async tag and set inferred span end time based on value

hghotra:
### What does this PR do?

Add sync/async tag and set inferred span end time based on value. 

### Motivation

Some upstream managed services like API Gateway invoke the Lambda function synchronously and wait for the function execution to finish. In this case, the inferred span for API Gateway should finish after the Lambda span finishes. 

In other cases such as SNS, SQS, etc., the inferred spans should finish when the Lambda execution starts.

### Testing Guidelines

**Sync Inferred Span**

image

**Async Inferred Span**

image

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/wrapper.py
```diff
@@ -201,10 +205,17 @@ def _after(self, event, context):
                 if status_code:
                     self.span.set_tag("http.status_code", status_code)
                 self.span.finish()
+
             if self.inferred_span:
+                # TODO FIXME: should we move this logic to
+                # the inferred span methods where this is relevant?
```
agocs, line 211: I think we don't know the status code until `_after`, and we don't know `self.span.start` until after the inferred span has been created. We could break this out into a `finish_inferred_span` method if you think that would be cleaner

agocs (approved):
LGTM. I left a comment on your comment

datadog_lambda/wrapper.py
hghotra: Okay yes, you're right we don't have the status code until `_after` (that's what this comment was referring to). The rest looks fine, I'm gonna go through and refactor the entire inferred spans logic sometime soon so I'll move this bit into a `_ finish_inferred_span` method.

PR #204, opened 2022-01-31
Inferred spans for AWS Managed Services

hghotra:
### What does this PR do?

Inferred Spans for the following AWS Managed Services:

- API Gateway V1
- API Gateway V2
- SQS
- SNS
- SNS => SQS
- Kinesis
- EventBridge (custom events)
- DynamoDB
- S3

### Testing Guidelines

Unit Tests:
`poetry run nose2`

Integration Tests:
`BUILD_LAYERS=true UPDATE_SNAPSHOTS=true DD_API_KEY=<api_key> aws-vault exec sandbox-account-admin -- ./scripts/run_integration_tests.sh`

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


README.md
```diff
@@ -103,6 +103,18 @@ Initialize the Datadog tracer when set to `true`. Defaults to `false`.
 
 Set to `true` to merge the X-Ray trace and the Datadog trace, when using both the X-Ray and Datadog tracing. Defaults to `false`.
 
+### DD_INFERRED_SPANS (experimental)
```
nhinsch, line 106: Just want to confirm that if you plan to immediately release this layer after merging. If not, I would break out the docs into another PR so they can be released later.

README.md
```diff
@@ -103,6 +103,18 @@ Initialize the Datadog tracer when set to `true`. Defaults to `false`.
...
+
+Inferred Spans are spans that Datadog can create based on incoming event metadata.
+Set `DD_INFERRED_SPANS` to `true` to infer spans based on Lambda events.
+Inferring upstream spans is only supported if you are using the [Datadog Lambda Extension](https://docs.datadoghq.com/serverless/libraries_integrations/extension/).
+Defaults to `false`.
+Infers spans for:
+
+- API Gateway REST events
+- API Gateway websocket events
+- HTTP API events
```
nhinsch, line 116: It looks like in the code there are a lot of other options as well (Kinesis, DynamoDB, etc.)

datadog_lambda/tracing.py
```diff
@@ -170,23 +185,98 @@ def extract_context_from_http_event_or_context(event, lambda_context):
...
-    except Exception:
+    except Exception as e:
+        logger.debug("The trace extractor returned with error %s", e)
+        return extract_context_from_lambda_context(lambda_context)
+
+
+def extract_context_from_eventbridge_event(event, lambda_context):
+    """
+    Extract datadog trace context from an EventBridge message's Details.
+    Details is often a weirdly escaped almost-JSON string. Here we have to correct for that.
```
nhinsch, line 241: This comment says that you need to extract the context from an almost-JSON like string, and that you have to correct for it, but I don't see anything in the function about extracting the context from a string or correcting any formatting --  am I missing something?

nhinsch:
For the integration test failures, for now you can add this line to fix them:
https://github.com/DataDog/datadog-agent/pull/10692/files#diff-9de960c72d49d9709578a28dcad0619f81979366f24602ee8a29502f2a5e5c3fR27

README.md
hghotra: Yes, the plan is to release this layer immediately. I've updated the docs with your suggestions, thanks!

datadog_lambda/tracing.py
hghotra: Yeah, I think Chris added that comment because of a bug in moto where `Details` is a JSON string that's not properly escaped & requires using something like `ast.literal_eval` rather than `json.loads()`. However, this is just an issue with the testing library, and not an actually in AWS, so I've updated the comment.

datadog_lambda/tracing.py
```diff
@@ -377,13 +480,394 @@ def set_dd_trace_py_root(trace_context_source, merge_xray_traces):
...
+        "detail_type": event["detail-type"],
+    }
+    InferredSpanInfo.set_tags(
+        tags,
+        synchronicity="async",
+        tag_source="self",
+    )
+    dt_format = "%Y-%m-%dT%H:%M:%SZ"
+    timestamp = event["time"]
+    dt = datetime.strptime(timestamp, dt_format)
```
astuyve, line 848: :/ I wish we had millisecond resolution here, but this will have to do for now

astuyve (approved):

datadog_lambda/tracing.py
hghotra: Yeah, we need a figure out a solution for this and potentially other spans down the road with a similar issue.

Issue #225, opened 2022-04-20
TypeError: argument of type 'NoneType' is not iterable


suwat-netprom:
## Expected Behavior
can show log not have issue

## Actual Behavior
`[ERROR] TypeError: argument of type 'NoneType' is not iterable
Traceback (most recent call last):
  File "/opt/python/lib/python3.8/site-packages/datadog_lambda/wrapper.py", line 142, in __call__
    self.response = self.func(event, context, **kwargs)
  File "/var/task/wsgi_handler.py", line 59, in handler
    if "_serverless-wsgi" in event:`

## Steps to Reproduce the Problem

  1. test call api with postman to api-gateway
  2. goto lambda -> cloudwatch

## Specifications

  - aws lambda
  - python flask framework
  - python flask_restful
  - serverless-python-requirements plugin
  - serverless-wsgi plugin
  - serverless-plugin-datadog plugin
  - Datadog Lambda Layer version: datadog-python38_3.57.0
  - Python version: 3.8
  
## Stacktrace
  
![image](https://user-images.githubusercontent.com/14134992/164277223-b543227a-53a5-4c9c-930f-a89ea252f58a.png)

![image](https://user-images.githubusercontent.com/14134992/164253312-851e374b-ef2d-49f9-a9b2-c2755b11b488.png)

![image](https://user-images.githubusercontent.com/14134992/164271676-96ca249c-6e1d-463d-a0ce-254baed0570e.png)

thany you

  
  Paste here
  ```

zARODz11z:
Hello, it seems the cause of the error is that the lambda event is empty (None). Any ideas why it might be set to None?

suwat-netprom:
I just started And I can't figure it out. I had never encountered an incident like this before.

astuyve:
Hi @suwat-netprom - looking at your report I suspect this library doesn't work with `serverless-wsgi`. This is because `serverless-wsgi` converts the incoming lambda event into a standard WSGI [event](https://github.com/logandk/serverless-wsgi#features).

We can of course fix this logic to not cause an error, but this library is effectively predicated on having the `event` object accessible, so without that I'm not sure we'll able to fundamentally help. We can consider this a feature request for now.

suwat-netprom:
hello, Thanks for checking if it's compatible. will be very helpful. 

ps. My English isn’t good.

PR #236, opened 2022-06-27
feat: Replace list with set

astuyve:
### What does this PR do?

We don't know what a user may bring to the event arg, as libraries ran before us (see #225) can rewrite it.
Although list responds to in, it wouldn't really let us access headers unless we knew the format of the list. But a set will, so this PR allows for a set instead, which does work with the underlying calls in extract_context_from_http_event_or_context
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


kimi-p (approved):
lgtm

PR #244, opened 2022-08-23
[beta] Python Profiling support

astuyve:
### What does this PR do?

This PR enables the (beta) use of the new serverless profiling capabilities of ddtrace-py in Lambda.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/wrapper.py
```diff
@@ -181,7 +187,9 @@ def _before(self, event, context):
                 )
             else:
                 set_correlation_ids()
-
+            if profiling_env_var == True and is_cold_start():
+                print("starting profiler in _before()")
```
maxday, line 191: 👀 not sure we want to keep this?

astuyve: Nope

datadog_lambda/wrapper.py
```diff
@@ -35,12 +35,16 @@
...
+from ddtrace.profiling import profiler
 
 logger = logging.getLogger(__name__)
 
 dd_capture_lambda_payload_enabled = (
     os.environ.get("DD_CAPTURE_LAMBDA_PAYLOAD", "false").lower() == "true"
 )
+profiling_env_var = os.environ.get("DD_PROFILING_ENABLED", "false").lower() == "true"
+service_env_var = os.environ.get("DD_SERVICE", "DefaultServiceName")
+env_env_var = os.environ.get("DD_ENV", "dev")
```
maxday, line 47: should we leave that empty if no env has been set? I'm not sure the default value is "dev" ?

astuyve: Ah yeah we can use `None` here.

maxday (approved):
LGTM! do we want to build and release this layer in sandbox to see the impact on cold start (when profiling is disabled) since the package size has grown a bit?

astuyve:
We ran three versions of the same handler code (simple hello world).
- without-profiling (version 60 of this library, no profiler code at all)
- with profiling (pre-release of 61 of this library, which imports the profiler and turns it on)
- with profiling but disabled (pre-release of 61 of this library, which has the larger filesize compared to without-profiling)

With profiling packaged and enabled, we see ~200ish ms cold start increase.
We see almost no cold start increase for including profiling, but not loading that code.
image
image

PR #256, opened 2022-10-26
authorizer trace context

joeyzhao2018:
### What does this PR do?

Fixes https://github.com/DataDog/serverless-plugin-datadog/issues/252 for Python
Ticket: [SLS-2499](https://datadoghq.atlassian.net/browse/SLS-2499),  [RFC](https://docs.google.com/document/d/1q4JJoM4Ng-xTRScKrdvfNnMZtKjmgkT6z9iBaUUpNUs)

### Motivation

This solution is essentially adding an extra inferred span that connects the current two traces, i.e. the authorizer function trace and the main function trace, into one trace.
apigateway authorizer

### Implementation
- When in the authorizer function, inject the necessary trace information and the timestamp information into the returned response
- When in the main function, extract the information injected during authorizer function execution and create an inferred span based on it.
- More Details and test evidence: [google doc link](https://docs.google.com/document/d/1sHypySSwJKWbWzIEreCgw8XNNE74y-G6UedH8A6dVi4/edit?usp=sharing)

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/constants.py
```diff
@@ -42,3 +42,11 @@ class XrayDaemon(object):
     XRAY_TRACE_ID_HEADER_NAME = "_X_AMZN_TRACE_ID"
     XRAY_DAEMON_ADDRESS = "AWS_XRAY_DAEMON_ADDRESS"
     FUNCTION_NAME_HEADER_NAME = "AWS_LAMBDA_FUNCTION_NAME"
+
+
+class Headers(object):
+    Parent_Span_Finish_Time = "x-datadog-parent-span-finish-time"
```
astuyve, line 48: Can you add a comment here and to `Authorizing_Request_Id` explaining why we need each?

datadog_lambda/tracing.py
```diff
@@ -616,14 +720,54 @@ def create_inferred_span_from_api_gateway_websocket_event(event, context):
...
+                                "integrationLatency", 0
+                            )
+                        )
+                    )
+                ) * 1e6
+                upstream_authorizer_span = insert_upstream_authorizer_span(
+                    args, tags, start_time_ns, finish_time_ns
+                )
+            except Exception as e:
+                traceback.print_exc()
```
astuyve, line 748: Do we still need this?

datadog_lambda/tracing.py
```diff
@@ -689,10 +874,20 @@ def create_inferred_span_from_http_api_event(event, context):
...
+        )
+        if injected_authorizer_data:
+            inferred_span_start_ns = int(
+                injected_authorizer_data.get(Headers.Parent_Span_Finish_Time)
+            )
+    inferred_span_start_ns = inferred_span_start_ns or request_time_epoch_ms * 1e6
     span = tracer.trace("aws.httpapi", **args)
     if span:
         span.set_tags(tags)
-    span.start = request_time_epoch / 1000
```
astuyve, line 695: Is moving this into the conditional also a bug fix? Along with fixing the request time epoch?

datadog_lambda/tracing.py
```diff
@@ -651,14 +795,55 @@ def create_inferred_span_from_api_gateway_event(event, context):
...
+            event, _EventSource(EventTypes.API_GATEWAY, EventSubtypes.API_GATEWAY)
+        )
+        if injected_authorizer_data:
+            try:
+                start_time_ns = int(
+                    injected_authorizer_data.get(Headers.Parent_Span_Finish_Time)
+                )
+                finish_time_ns = (
+                    request_time_epoch_ms
+                    + (
```
astuyve, line 811: This block is pretty complex, can we extract it into a method?

datadog_lambda/tracing.py
```diff
@@ -689,10 +874,20 @@ def create_inferred_span_from_http_api_event(event, context):
         "span_type": "http",
     }
     tracer.set_tags({"_dd.origin": "lambda"})
+    inferred_span_start_ns = None
+    if decode_authorizer_context:
```
astuyve, line 878: Can we break this up into a method to simplify this conditional?
Then we could do `if injected_authorizer_data` here, and in the method, we could check if we should decode context.

datadog_lambda/tracing.py
```diff
@@ -689,10 +874,20 @@ def create_inferred_span_from_http_api_event(event, context):
...
+        injected_authorizer_data = get_injected_authorizer_data(
+            event, _EventSource(EventTypes.API_GATEWAY, EventSubtypes.HTTP_API)
+        )
+        if injected_authorizer_data:
+            inferred_span_start_ns = int(
+                injected_authorizer_data.get(Headers.Parent_Span_Finish_Time)
+            )
+    inferred_span_start_ns = inferred_span_start_ns or request_time_epoch_ms * 1e6
```
astuyve, line 886: we're using 1e6 in a couple of places in this PR, although I know we're converting ms to ns, and your variables are named perfectly, it might be helpful to extract this into a constant? `MILLIS_TO_NANOS` maybe? There probably is one in the time package.

datadog_lambda/tracing.py
```diff
@@ -973,14 +1168,39 @@ class InferredSpanInfo(object):
...
+        synchronicity: Optional[Literal["sync", "async"]] = None,
+        tag_source: Optional[Literal["labmda", "self"]] = None,
+    ):
+        if synchronicity is not None:
+            span.set_tags({InferredSpanInfo.SYNCHRONICITY: synchronicity})
+        if tag_source is not None:
+            span.set_tags({InferredSpanInfo.TAG_SOURCE: str(tag_source)})
+
+    @staticmethod
+    def is_async(span: Span) -> bool:
```
astuyve, line 1194: Good refactor here

datadog_lambda/wrapper.py
```diff
@@ -234,6 +263,26 @@ def _after(self, event, context):
...
+                self.encode_authorizer_context
+                and self.response
+                and self.response.get("principalId")
+                and self.response.get("policyDocument")
+            ):
+                # the finish_time_ns should be set as the end of the inferred span if it exist
+                #  or the end of the current span
+                reference_span = self.inferred_span if self.inferred_span else self.span
+                finish_time_ns = (
+                    reference_span.start_ns + reference_span.duration_ns
```
astuyve, line 277: Maybe this could be another method? Although I realize it's pretty python-y already. Up to you.

astuyve (approved):
Couple small notes, otherwise great work here Joey!

datadog_lambda/tracing.py
joeyzhao2018: Yes, assuming the conditional makes sense. However, after taking a closer look, `tracer.trace()` should always return a span. So I'll actually remove this `if span:` line.

datadog_lambda/tracing.py
joeyzhao2018: I further cleaned up this a bit. However, I do want keep the `if decode_authorizer_context:` at this scope so that we can see if it's False we will just skip completely.

datadog_lambda/tracing.py
joeyzhao2018: I tried to first use a const `MILLIS_TO_NANOS = 1e6`. I didn't like it because it lacks the operator. It should be more like `MILLIS_TO_NANOS(x) :  1e6 * x`.
Then I tried to define a function `millis_to_nano(x)`. But i feel using a function is a bit overkill and tends to hide information.
I think I'll keep it explicit like this for now.

datadog_lambda/wrapper.py
joeyzhao2018: I tried to refactor then I realized that the `finish_time_ns` calculation requires that the `reference_span` is finished already. 
So I'll skip the refactoring to save the `reference_span already finished` check. I'll add a comment instead.

joeyzhao2018: On a second thought, I decided to add the `reference_span already finished` check.

duncanista (approved):
Looking really good, nice job. Thanks for refactoring! 🚀

PR #260, opened 2022-10-31
feat(ddtrace): disable telemetry collection by default

mabdinur:
### What does this PR do?

Disables `ddtrace` instrumentation telemetry data collection by setting the following environment variable: `DD_INSTRUMENTATION_TELEMETRY_ENABLED=True`. This change is required to use `ddtrace>=1.5` in serverless environments to fix a performance regression in cold starts. 

Where this feature is enabled in the ddtrace library: https://github.com/DataDog/dd-trace-py/blob/v1.5.3/ddtrace/internal/writer.py#L514

### Motivation

Instrumentation Telemetry is designed to collect, transform and visualize usage, diagnostics and onboarding data for the various APM client libraries (tracer/profiler/debugger/appsec/etc.). The data collected and sent to the datadog backend contains information about an application (dependencies, traced libraries, python versions and implementations, etc.), host/environment data (os, hostname, kernel, container id, etc.) and ddtrace configurations. Collecting this data in lambda functions can increase cold starts by up to 2 seconds and the overhead of the tracer (initial tests show a 5-10% increase but this needs further verification).

Note - Enabling instrumentation telemetry is required for application security products.

### Testing 

Use the changes in this branch and `ddtrace>=1.5` to verify the performance overhead of tracing lambda functions. Test hot and cold starts with the following cases:
   - case 1 (baseline): `DD_INSTRUMENTATION_TELEMETRY_ENABLED=False`. 
      - expected result: Performance is similar to ddtrace v1.4
  - case 2: `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is not set. 
      - expected result: Same as case 1
  - case 3: `DD_INSTRUMENTATION_TELEMETRY_ENABLED=True`. 
      - expected result: Some additional overhead. Add results to this PR description. 

If the results from `case 3` are acceptable then this PR should be closed. If the additional overhead is not acceptable then we should move forward with this change. 

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -968,6 +968,20 @@ def mark_trace_as_error_for_5xx_responses(context, status_code, span):
             span.error = 1
 
 
+def disable_instrumentation_telemetry_by_default():
+    """
+    As of ddtrace v1.5 instrumentation telemetry collection is enabled by default.
+    This feature can add 5-10% overhead to the tracer and increase cold start.
+    Setting ``DD_INSTRUMENTATION_TELEMETRY_ENABLED=False`` will disable this feature.
+
+    Instrumentation Telemetry Collection is enabled when the first trace is sent to the
+    Agent. This feature must be disabled before tracing is configured.
+    """
+
+    if "DD_INSTRUMENTATION_TELEMETRY_ENABLED" not in os.environ:
+        os.environ["DD_INSTRUMENTATION_TELEMETRY_ENABLED"] = "false"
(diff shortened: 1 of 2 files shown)


astuyve:
sorry friends, I didn't realize I missed this in my notifications. Do we still need this?

astuyve:
Closing for now, will reopen when we determine cost/benefit for this.

PR #325, opened 2023-04-17
fix: Ensure authorized span headers are injected as `str`

Wesitos:
### What does this PR do?

It injects the authorizer span headers as a regular string instead of a byte string.

### Motivation

The Python 3.7 lambda runtime can't serialize responses that contain byte strings.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -219,7 +219,7 @@ class _LambdaDecorator(object):
         injected_headers[Headers.Parent_Span_Finish_Time] = finish_time_ns
         if request_id is not None:
             injected_headers[Headers.Authorizing_Request_Id] = request_id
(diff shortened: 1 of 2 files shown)


joeyzhao2018 (commented):
Putting request changes here to block. We had to use encoding because
> Injecting extra data to the original response can break invocations using cached authorization data. Symptoms: The first (authorizing) invocation succeeded. But the following requests received Internal Server Error messages.

Therefore, TODO on our side: we need to test the encode and decode case here. We should add the case to the the integration test.

Hi @Wesitos, could you please open an issue on the serialization problem mentioned in the description here so that we can have more information? Thank you in advance.

Issue #329, opened 2023-05-05
"Unable to import module 'datadog_lambda.handler': cannot import name 'DEFAULT_CIPHERS'"


pmaurais:
## Expected Behavior
The lambda runs and logs to DD

## Actual Behavior
``` "Unable to import module 'datadog_lambda.handler': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/var/task/urllib3/util/ssl_.py)", ```

Likely related to https://github.com/urllib3/urllib3/issues/2168

## Steps to Reproduce the Problem

  1. Create a SAM template that uses the latest version of the DD Python Extension
  2. Create a Lambda that uses python 3.9
  3. sam build -u, sam deploy
  4. Attempt to execute the lambda 

## Specifications

  - Datadog Lambda Layer version: 43
  - Python version: 74
  
## Stacktrace
  
  ```
 [ERROR] Runtime.ImportModuleError: Unable to import module 'datadog_lambda.handler': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/var/task/urllib3/util/ssl_.py)
Traceback (most recent call last):
  ```

**Comment by pmaurais** [comment]
@kid-92

**Comment by filipponova** [comment]
Same error here.

Python 3.10

```bash
Runtime.ImportModuleError: Unable to import module 'datadog_lambda.handler': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/var/task/urllib3/util/ssl_.py)
```

It was working until last week. :/

filipponova:
In Python 3.8 a different error with link to this issue: https://github.com/urllib3/urllib3/issues/2168

```bash
Runtime.ImportModuleError: Unable to import module 'datadog_lambda.handler': urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2k-fips  26 Jan 2017. See: https://github.com/urllib3/urllib3/issues/2168
```

astuyve:
Hey folks, thanks for reaching out! We're investigating this and will get back to you

astuyve:
Hi @filipponova and @pmaurais - I see you mentioned this was working until last week. Did you upload from an earlier version of this library (Datadog Lambda Python)? The most recent layer version is 74, was that working until recently?
The Datadog Lambda Extension is written in Go, and that seems unrelated here.

[...]

PR #333, opened 2023-05-16
DD_SERVICE_MAPPING implementation with tests

zARODz11z:
### What does this PR do?
allows users to remap the service name on inferred spans created in dd lambda js. Pairs with https://github.com/DataDog/documentation/pull/18144

### Motivation
SLS-3501

### Testing Guidelines
unit

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


pyproject.toml
```diff
@@ -34,7 +34,7 @@ requests = { version ="^2.22.0", optional = true }
 nose2 = { version= "^0.9.1", optional = true }
 flake8 = { version = "^3.7.9", optional = true }
 httpretty = {version = "^0.9.7", optional = true }
-
+urllib3 = "<2.0.0"
```
zARODz11z, line 37: due to

https://github.com/DataDog/datadog-lambda-python/issues/329

datadog_lambda/tracing.py
```diff
@@ -685,9 +685,32 @@ def create_inferred_span(
...
+            mapping[key.strip()] = value.strip()
+
+    return mapping.get(service_name)
+
+
 def create_inferred_span_from_lambda_function_url_event(event, context):
     request_context = event.get("requestContext")
+    api_id = request_context.get("apiId")
     domain = request_context.get("domainName")
+    # Attempt to get the service mapping for the domain so that we don't remap all apigw services
```
astuyve, line 705: Hmm, this complexity is necessary but I wonder if we can formalize it a bit more?

Right now a bug could be inadvertently introduced in line 707 or 710, which would `fail through` to the base case of `domain`.

That's good behavior, but when we see logic on different lines our minds don't always connect them. Would a switch statement help make the causality clear? Wdyt?

zARODz11z: perhaps 
```
api_id = request_context.get("apiId")
domain = request_context.get("domainName")

# Attempt to get the service mapping for the domain so that we don't remap all apigw services
# Allows the customer to have more fine-grained control
apiid_mapping = get_service_mapping(str(api_id))

lambda_url_mapping = None
if not apiid_mapping:
    # If the domain mapping is not found,
    # attempt to get the service mapping for 'lambda_api_gateway'
    lambda_url_mapping = get_service_mapping("lambda_url")

# If neither mapping is found, default to the domain name
if apiid_mapping:
    service_name = apiid_mapping
elif lambda_url_mapping:
    service_name = lambda_url_mapping
else:
    service_name = domain
```

tests/test_tracing.py
```diff
@@ -732,6 +734,459 @@ def _basic_common_checks(
         self.assertEqual(span.get_tag("request_id"), "abc123")
 
 
+class TestServiceMapping(unittest.TestCase):
```
astuyve, line 737: Really good job on this battery of tests. Fall-through config is important but also behavior is often left untested and results in latent bugs months later, and these seem very thorough.

datadog_lambda/tracing.py
zARODz11z:
```
api_id = request_context.get("apiId")
domain = request_context.get("domainName")

# Attempt to get the service mapping for the domain so that we don't remap all apigw services
# Allows the customer to have more fine-grained control
apiid_mapping = get_service_mapping(str(api_id))

# Attempt to get the service mapping for 'lambda_api_gateway'
lambda_url_mapping = get_service_mapping("lambda_url")

if apiid_mapping:
    service_name = apiid_mapping
elif lambda_url_mapping:
    service_name = lambda_url_mapping
else:
    # If neither mapping is found, default to the domain name
    service_name = domain

```

datadog_lambda/tracing.py
```diff
@@ -685,9 +685,32 @@ def create_inferred_span(
     return None
 
 
+def get_service_mapping(service_name):
```
astuyve, line 688:
```suggestion
service_mapping = {}
for entry in os.getenv("DD_SERVICE_MAPPING", "").split(","):
    parts = entry.split(":")
    if len(parts) == 2:
        key, value = parts
        service_mapping[key.strip()] = value.strip()
```

datadog_lambda/tracing.py
astuyve: Okay good start - definitely clearer. Let’s define the requirements in plain english real quickly:
Highest priority - user overrides the specific function URL.
Medium priority - user overrides lambda_url
Lowest priority - no customization

To me, this means we should have a pretty clean conditional, and we can avoid mutating a variable unnecessary by refactoring this to its own function (don't actually do this, but for example):
```python
def function_url_mapping(api_id_string):
  if get_service_mapping(api_id_string):
    return get_service_mapping(api_id)
  else if get_service_mapping("lambda_url"):
    return get_service_mapping("lambda_url")
   else:
     return domain
```

But this has some problems. Firstly, `get_service_mapping` re-parses the ENV var with every call, which we only need to do once when the function starts. We don't need a function for this, we could simply make python do this when the class loads. This means refactoring things a bit. I suggested above that we simply use a dict to track the mapping, which now lets us make our work more pythonic.

```python
def function_url_mapping(api_id_string):
  if api_id_string in service_mapping:
    return service_mapping.get(api_id_string)
  else if "lambda_url" in service_mapping:
    return service_mapping.get("lambda_url")
  else:
    return domain
```

How does that read to you? Does that feel more like the business rules we laid out above?

astuyve (changes requested):

datadog_lambda/tracing.py
zARODz11z: 100% this looks awesome! I like the added touch of parsing the env var as the class loads, makes it really clean

datadog_lambda/tracing.py
zARODz11z: I wonder if making this a global variable has any implications in production? basically if we make it a global variable then that service_mapping dict will only get assigned once the whole tracing.py module/file is loaded.

I know it broke my tests because the DD_SERVICE_MAPPING env var isn't being parsed for each test, only for the 1st one

zARODz11z: yeah I think its best to keep this as a function

zARODz11z: or else it becomes hard to test without reloading the tracing.py module

datadog_lambda/tracing.py
zARODz11z: this becomes hard to test though as DD_SERVICE_MAPPING will only be parsed once, when the module loads

tests/test_tracing.py
```diff
@@ -14,22 +16,7 @@
...
-    mark_trace_as_error_for_5xx_responses,
-    set_correlation_ids,
-    set_dd_trace_py_root,
-    _convert_xray_trace_id,
-    _convert_xray_entity_id,
-    _convert_xray_sampling,
-    InferredSpanInfo,
-    extract_context_from_eventbridge_event,
-)
+import datadog_lambda.tracing as dt
```
zARODz11z, line 19: did this so we can importlib.reload the module, to force the parsing of DD_SERVICE_MAPPING for each test. (kinda a hacky way of doing this but it fixes the tests and lets us stay with the logic of only parsing DD_SERVICE_MAPPING when the tracing.py module is loaded.

zARODz11z: I fixed all other references with find and replace with the dt alias

tests/test_tracing.py
```diff
@@ -732,6 +727,477 @@ def _basic_common_checks(
...
+        # Testing the second event
+        event2 = copy.deepcopy(
+            original_event
+        )  # Create a deep copy so we don't modify the original event
+        # Modify event2 as necessary
+        event2["requestContext"]["apiId"] = "different"
+        span2 = dt.create_inferred_span(event2, ctx)
+        self.assertEqual(span2.get_tag("operation_name"), "aws.apigateway.rest")
+        self.assertEqual(
+            span2.service, "70ixmpl4fl.execute-api.us-east-2.amazonaws.com"
```
zARODz11z, line 787: default case asserted :)

tests/test_tracing.py
zARODz11z: thank you! `Fall-through config is important` not sure if you are implying I should add tests?

joeyzhao2018 (approved):
LGTM with some nitpicks

datadog_lambda/tracing.py
```diff
@@ -685,9 +685,44 @@ def create_inferred_span(
...
+    for entry in val.split(","):
+        parts = entry.split(":")
+        if len(parts) == 2:
+            key = parts[0].strip()
+            value = parts[1].strip()
+            new_service_mapping[key] = value
+    return new_service_mapping
+
+
+def set_service_mapping(new_service_mapping):
```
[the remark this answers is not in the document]
duncanista: Wouldn't it be better to rename them with a prefix underscored to not expose them @joeyzhao2018 ?

So `set_service_mapping` => `_set_service_mapping` making it private. Same for `get_service_mapping`.

tests/event_samples/lambda-url.json
```diff
@@ -0,0 +1,47 @@
+{
```
duncanista, line 1: Where does this file come from?

datadog_lambda/tracing.py
```diff
@@ -685,9 +685,44 @@ def create_inferred_span(
     return None
 
 
+def create_service_mapping(val):
+    new_service_mapping = {}
+    for entry in val.split(","):
+        parts = entry.split(":")
+        if len(parts) == 2:
```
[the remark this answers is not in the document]
zARODz11z: added a check for that here and added a test case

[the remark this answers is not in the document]
zARODz11z: https://github.com/DataDog/datadog-lambda-python/pull/333/files#diff-1f82cebfdbd1dd410514f8a66ff6d3062c390662e153ff381218eb9492d64796R750

tests/event_samples/lambda-url.json
zARODz11z: copied from [datadog-lambda-js](https://github.com/DataDog/datadog-lambda-js/blob/50a1b243aa89c973e7f7b5232f4529b312f7bc73/event_samples/lambda-function-urls.json) I wanted a test case for lambda-url and datadog-lambda-python didnt have this event sample

duncanista: Nice, great work!

datadog_lambda/tracing.py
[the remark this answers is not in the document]
zARODz11z: good point, I moved it and edited the tests

duncanista (approved):

astuyve (approved):

PR #337, opened 2023-05-23
feat: add `peer.service`

duncanista:
### What does this PR do?

Adds the field peer.service to inferred spans.
The field is set to the current process environment variable DD_SERVICE.

### Motivation

Add another dimension to services in APM.

### Testing Guidelines

- Integration tests with a service to check that inferred spans now have the tag.

### Additional Notes

Refactoring in `tracing.py` is needed.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tracing.py
```diff
@@ -43,9 +43,15 @@
 
 logger = logging.getLogger(__name__)
 
+AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME"
+DD_SERVICE = "DD_SERVICE"
```
zARODz11z, line 47: Where is AWS_LAMBDA_FUNCTION_NAME documented? I couldn't find it on https://docs.datadoghq.com/serverless/aws_lambda/configuration/?tab=datadogcli which is in https://github.com/DataDog/datadog-lambda-python/blob/6f74908584dc8a937cfe026b4807b4a71dffc08e/README.md#L17

zARODz11z: I see DD_SERVICE mentioned there

datadog_lambda/tracing.py
```diff
@@ -43,9 +43,15 @@
...
+
 dd_trace_context = {}
 dd_tracing_enabled = os.environ.get("DD_TRACE_ENABLED", "false").lower() == "true"
 
+function_env_var = os.environ.get(AWS_LAMBDA_FUNCTION_NAME, "function")
+service_env_var = os.environ.get(DD_SERVICE, function_env_var)
+
```
zARODz11z, line 54: Just so I understand peer.service tag is resolution is (in the order of priority)
1. Val of DD_SERVICE
2. Val of AWS_LAMBDA_FUNCTION_NAME
3. "function"

I didn't read much of the peer.service rfc but again i'd like to know more about AWS_LAMBDA_FUNCTION_NAME

zARODz11z:
What about adding peer.service on the function execution span?
https://github.com/DataDog/datadog-lambda-python/pull/337/files#diff-eb1f51efb218a7a829e913e4ff6b17ce27be8a777bc20c283e5b843e5db151efR1177

tests/test_tracing.py
```diff
@@ -805,6 +805,7 @@ def test_create_inferred_span_from_api_gateway_event(self):
         self.assertEqual(span.span_type, "http")
         self.assertEqual(span.get_tag(InferredSpanInfo.TAG_SOURCE), "self")
         self.assertEqual(span.get_tag(InferredSpanInfo.SYNCHRONICITY), "sync")
+        self.assertEqual(span.get_tag("peer.service"), "function")
```
zARODz11z, line 808: I think it would be good to have standalone tests also, for example here you are assertion just the fallback. I know the implementation is simple but as I mentioned here https://github.com/DataDog/datadog-lambda-python/pull/337#discussion_r1204642506 there seems different resolutions of peer.service

datadog_lambda/tracing.py
duncanista: I just did this due to the profiler in Python, not really the best option, but it is what we have.

tests/test_tracing.py
duncanista: Yes, I know.
Nonetheless, since I cannot just set `DD_SERVICE` in `setUp` because `tracing.py` is initializing the variable that stores it. There is no way of setting a new value in the actual state of `tracing.py`. I want to change this in another PR where we refactor the file and how inferred spans are handled.

datadog_lambda/tracing.py
duncanista: As per our conversation, https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

duncanista:
> What about adding peer.service on the function execution span?
https://github.com/DataDog/datadog-lambda-python/pull/337/files#diff-eb1f51efb218a7a829e913e4ff6b17ce27be8a777bc20c283e5b843e5db151efR1177

This would have to be handled by `dd-trace-py` since we are doing upstream only in this PR.

datadog_lambda/tracing.py
zARODz11z: Hmm something isn't clicking, how does this relate to the profiler?

zARODz11z:
> > What about adding peer.service on the function execution span?
> https://github.com/DataDog/datadog-lambda-python/pull/337/files#diff-eb1f51efb218a7a829e913e4ff6b17ce27be8a777bc20c283e5b843e5db151efR1177
> 
> This would have to be handled by `dd-trace-py` since we are doing upstream only in this PR.

Interesting, so you're saying in the tracer we'd add peer.service on the actual function execution span? Doing this in a different pr is fine with me but what you said doesn't make sense to me since the function execution span is created here so why would we want to add the peer.service tag in the tracer instead of here in dd lambda py? Unless it's because any downstream spans in relation to the function execution span are created in the tracer thus I'd have the appropriate value of what the peer.service tag should equal

I.e. if we have a function that makes an aws sdk request to sqs then because that aws sdk request span is created in the tracer we would add another tag to the function execution span pointing to the sqs queue???

zARODz11z:
I wonder how we would edit the function execution span in the tracer. For some reason that isn't clicking, I assume we start and finish it in dd lambda py so adding that peer.service span in the tracer after its been finished seems like it may be difficult but I've never done that. Lmk

datadog_lambda/tracing.py
```diff
@@ -801,6 +808,7 @@ def create_inferred_span_from_api_gateway_websocket_event(
         "connection_id": request_context.get("connectionId"),
         "event_type": request_context.get("eventType"),
         "message_direction": request_context.get("messageDirection"),
+        "peer.service": service_env_var,
```
joeyzhao2018, line 811: 💡  just a general comment and it doesn't have to be done in this pr: we should create a `general_tags()` method and save some repetition.

duncanista: I agree, we should refactor this, but it's a really long file. We should follow a similar approach in Node.

datadog_lambda/wrapper.py
duncanista, line 1: I think the logic in this file should not be handled by this PR.

datadog_lambda/tracing.py
```diff
@@ -43,9 +43,15 @@
...
 logger = logging.getLogger(__name__)
 
+AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME"
+DD_SERVICE = "DD_SERVICE"
+
 dd_trace_context = {}
 dd_tracing_enabled = os.environ.get("DD_TRACE_ENABLED", "false").lower() == "true"
 
+function_env_var = os.environ.get(AWS_LAMBDA_FUNCTION_NAME, "function")
+service_env_var = os.environ.get(DD_SERVICE, function_env_var)
```
duncanista, line 53: The fallback will be empty `""` in order to keep consistency with JS.

duncanista: Fallback changed to `None`.

PR #343, opened 2023-06-23
Aj/proactive init

astuyve:
### What does this PR do?

This adds support for tagging proactively initialized lambda functions.

### Testing Guidelines

I built a pre-release and tested it, verifying we're correctly tagging proactively initialized lambda functions
image

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/cold_start.py
+++ b/datadog_lambda/cold_start.py
@@ -20,7 +20,7 @@ def set_cold_start(init_timestamp_ns):
     global _proactive_initialization
     if not _lambda_container_initialized:
         now = time.time_ns()
-        if (now - init_timestamp_ns) // 1_000_000_000:
+        if (now - init_timestamp_ns) // 1_000_000_000 > 10:
             _cold_start = False
             _proactive_initialization = True
         else:
--- a/datadog_lambda/cold_start.py
+++ b/datadog_lambda/cold_start.py
@@ -6,17 +6,28 @@ import logging
 logger = logging.getLogger(__name__)
 
 _cold_start = True
+_proactive_initialization = False
 _lambda_container_initialized = False
 
 
-def set_cold_start():
+def set_cold_start(init_timestamp_ns):
     """Set the value of the cold start global
 
     This should be executed once per Lambda execution before the execution
     """
     global _cold_start
     global _lambda_container_initialized
-    _cold_start = not _lambda_container_initialized
+    global _proactive_initialization
+    if not _lambda_container_initialized:
+        now = time.time_ns()
+        if (now - init_timestamp_ns) // 1_000_000_000:
+            _cold_start = False
+            _proactive_initialization = True
+        else:
+            _cold_start = not _lambda_container_initialized
+    else:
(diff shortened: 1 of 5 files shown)


joeyzhao2018 (approved):
LGTM

datadog_lambda/cold_start.py
```diff
@@ -6,17 +6,28 @@
...
     This should be executed once per Lambda execution before the execution
     """
     global _cold_start
     global _lambda_container_initialized
-    _cold_start = not _lambda_container_initialized
+    global _proactive_initialization
+    if not _lambda_container_initialized:
+        now = time.time_ns()
+        if (now - init_timestamp_ns) // 1_000_000_000 > 10:
+            _cold_start = False
```
pvicente, line 24: Hey @astuyve @joeyzhao2018 should it also be `_cold_start = not _lambda_container_initialized`? I'm getting use cold_start stats from apms and with this change I will need to take into account new `proactive_initialization= true` to measure it like before.

Based on [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html#:~:text=When%20you%20use,using%20provisioned%20concurrency.) I'm not sure if in this situation cold_start should be false?

Any thoughts?

Thanks

astuyve: Hi @pvicente - We're actively working with AWS on documentation for this and I can't say more right now, but you should open a support ticket with them and tell them AJ Stuyvenberg sent you ;)

A bit more about what I can say:
With this change you'll still get `cold_start:true` if the AWS Lambda function actually experienced a cold start, like so:
image

However if the function is proactively initialized, then it's not a cold start in the sense that there was no additional latency caused by sandbox initialization (as init occurred seconds, hours, or minutes before the first request was served by the sandbox).

We now tag those correctly, so when we use APM to select this python invocation:
image
We can see that `initialization` occurred 3m59s before the function was executed:
image


I hope I've helped!

pvicente: Thanks for the explanation @astuyve 🥇

datadog_lambda/cold_start.py
```diff
@@ -6,17 +6,28 @@
...
+            _proactive_initialization = True
+        else:
+            _cold_start = not _lambda_container_initialized
```
pvicente, line 27: @astuyve I think on this case `_proactive_initialization = False` see my case from APM stats:

Screenshot 2023-07-07 at 14 28 28

astuyve: We initialize `_proactive_initialization` to `False` on line 9 so we shouldn't need to redeclare that here.

In this case rolling up to `top 10` is skewing the results:
image
I can see something similar in one of my orgs:
image
But when I remove the `pie chart` grouping, I see many `cold_start: true` spans:
image

However for clarity, line 27 should always be `false` as we're already inside of a conditional checking for `_lambda_container_initialized == false`

PR #358, opened 2023-08-04
feat: EventBridge to SQS support

duncanista:
### What does this PR do?

Adds support for tracing <EventBridge to SQS> payloads.

### Motivation

It's possible to have distributed tracing on it, so why not?

### Testing Guidelines

Added unit tests.
Also tested manually.

Screenshot 2023-08-04 at 8 20 53 PM

### Additional Notes

Since EventBridge doesn't provide resolution in milliseconds, we end up with really large spans for it.
Maybe this can be fixed by adding a new field. Research is still needed for this.

Screenshot 2023-08-04 at 8 21 17 PM

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -236,7 +236,8 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     # EventBrdige => SQS
     try:
-        return _extract_context_from_eventbridge_sqs_event(event)
+        trace_id, parent_id, sampling_priority = _extract_context_from_eventbridge_sqs_event(event)
+        return trace_id, parent_id, sampling_priority
     except Exception:
         logger.debug("Failed extracting context as EventBridge to SQS.")
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -233,6 +233,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     Falls back to lambda context if no trace data is found in the SQS message attributes.
     """
+
+    # EventBrdige => SQS
+    try:
+        return _extract_context_from_eventbridge_sqs_event(event)
+    except Exception:
+        logger.debug("Failed extracting context as EventBridge to SQS.")
+
     try:
         first_record = event.get("Records")[0]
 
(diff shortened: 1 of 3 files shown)


datadog_lambda/tracing.py
```diff
@@ -233,6 +233,18 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     Falls back to lambda context if no trace data is found in the SQS message attributes.
     """
+
```
astuyve, line 236: We should change the title of this method

datadog_lambda/tracing.py
```diff
@@ -283,6 +295,30 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
         return extract_context_from_lambda_context(lambda_context)
 
 
+def _extract_context_from_eventbridge_sqs_event(event):
```
astuyve, line 298: Shouldn't this be the same as the `_extract_context_from_eventbridge_event`, only with a different body/input?

astuyve (approved):
Couple notes, but functionally this looks good!

datadog_lambda/tracing.py
duncanista: I do agree. Nonetheless, if I used that method and failed, which is largely expected, it will directly try to extract the context from the lambda context. Instead of throwing an error so the next use case, `SNS => SQS`, is evaluated.

What do you think of this current approach? I would prefer rewriting the logic of the methods involved. Although I fear a refactor might take more pieces in account.

datadog_lambda/tracing.py
duncanista: Since we are including EventBridge? or what name would be more clear?

I think the current naming is adecuate since we are extracting context from either SNS or SQS.

`SNS => SQS` and `EB => SQS` are variants of it.

If you ask me, I'd refactor the whole method to separate SNS and SQS. Due to how they change between the type and value.

datadog_lambda/tracing.py
astuyve: This approach is good for now, thanks for the clarification!

datadog_lambda/tracing.py
astuyve: yeah probably better to refactor, but let's just rename for now.

datadog_lambda/tracing.py
duncanista: Definitely a note here to refactor the whole inferred spans.

PR #362, opened 2023-08-22
change the traceback format to print in one line

joeyzhao2018:
### What does this PR do?

Format the traceback printout to one line

### Motivation

Otherwise the printout would be split into multiple lines and would be hard to read and search.

### Testing Guidelines

This trick is just to replace `\n` with `\r` and the result looks like this:

Before:
Screenshot 2023-08-22 at 10 03 38 AM

After:
Screenshot 2023-08-22 at 10 04 15 AM

If we replace with `\n` with `\\n` instead:
Screenshot 2023-08-22 at 10 49 26 AM

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -126,7 +126,9 @@ class _LambdaDecorator(object):
                 logger.debug("datadog_lambda_wrapper already wrapped")
                 return _NoopDecorator(func)
         except Exception as e:
-            logger.error("Error %s. Traceback: %s", e, traceback.format_exc().replace("\n", "\r"))
+            logger.error(
+                "Error %s. Traceback: %s", e, traceback.format_exc().replace("\n", "\r")
+            )
             return func
 
     def __init__(self, func):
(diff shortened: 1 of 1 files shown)


astuyve (approved):

datadog_lambda/wrapper.py
```diff
@@ -204,8 +206,10 @@ def __init__(self, func):
             patch_all()
 
             logger.debug("datadog_lambda_wrapper initialized")
-        except Exception:
-            traceback.print_exc()
+        except Exception as e:
+            logger.error(
+                "Error %s. Traceback: %s", e, traceback.format_exc().replace("\n", "\r")
```
astuyve, line 211: One minor nit, we might consider splitting this out into a small util method, as we've got it in a few places and there doesn't seem to be something we'd change depending on input. Maybe the message? Even that can be a param.

joeyzhao2018: been thinking about this. will do.

datadog_lambda/wrapper.py
```diff
@@ -125,8 +125,10 @@ def __new__(cls, func):
             else:
                 logger.debug("datadog_lambda_wrapper already wrapped")
                 return _NoopDecorator(func)
-        except Exception:
-            traceback.print_exc()
+        except Exception as e:
+            logger.error(
+                "Error %s. Traceback: %s", e, traceback.format_exc().replace("\n", "\r")
```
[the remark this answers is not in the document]
joeyzhao2018: i don't think so. it seems that python just uses `\n`. i saw some people replace it with `\\n`. i added a screen shot in the description to show the difference.

PR #380, opened 2023-10-17
feat: Support urllib3 <= 2.1 for py 3.11

### LINKED ISSUE #376: URLLib3 is pinned below version 2
## Expected Behavior

Stable dependencies that don't downgrade without explanation.

## Actual Behavior

Urllib3 downgrades from 2.x after installing datadog-lambda

## Steps to Reproduce the Problem

  1. poetry init
  2. poetry add boto3
  3. poetry install
  4. poetry lock
  5. poetry show | grep urllib
  6. poetry add datadog-lambda
  7. poetry install
  8. poetry lock
  9. poetry show | grep urllib

## Specifications

The below details are provided as asked for, but are irrelevant

  - Datadog Lambda Layer version: N/A
  - Python version: 3.11
  
## Stacktrace
  
N/A; this bug is in the packaging 🙃

astuyve:
### What does this PR do?
Followup to #329
Fixes #376

### Motivation
Right now boto3 is at 1.26.90 for 3.7, 3.8, 3.9, and 3.10. This causes a known issue when users supply version >= 2.0 of [urllib3](https://urllib3.readthedocs.io/en/latest/v2-migration-guide.html#importerror-cannot-import-name-default-ciphers-from-urllib3-util-ssl).

Previously we pinned urllib3 to all versions of this library, but we'll open this up for Python 3.11.

Also, this marks urllib3 as optional as this library doesn't directly depend on it.

### Testing Guidelines
I tested importing boto3 using urllib3==2.0.7 on python 3.10, it fails to import.
I tested the same function code w/ urllib3==2.0.7 on python 3.11 and the code runes fine.

This pin and the automated tests all pass, so we should be confident this change will open up py3.11 users to urllib3 2.0 without impacting other users using incompatible runtimes (3.7, 3.8, 3.9, and 3.10)

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


LewisCowlesMotive:
👍  ship it 🙏 🚀

duncanista (approved):
Thanks for fixing this!

LGTM – just one question.

Could you be more specific on
> I tested this on 3.11 and see it functions correctly, but importing boto3 in 3.10 and under fails.

I'm not sure I understand it correctly.

astuyve:
Sorry, I'll clarify and update:
I tested importing boto3 using urllib3==2.0.7 on python 3.10, it fails to import.
I tested the same function code w/ urllib3==2.0.7 on python 3.11 and the code runes fine.

LewisCowlesMotive:
To confirm (for me please), 

```
boto3 = { version = "^1.28.0", optional = true }
```

The only material difference from my PR, makes this work with Python 3.10, or it is still broken if it selects boto3?

astuyve:
Yup, just 3.11

LewisCowlesMotive:
Works for me, it's the only one I'm using 😄

PR #405, opened 2023-12-13
Use one connection for all communication with extension.

purple4reina:
### What does this PR do?

Changes all communication with the extension to using a persistent http connection.

### Motivation

This change reduces runtime duration by somewhere between 5-10%.

Screenshot 2023-12-14 at 10 01 45 AM

### Testing Guidelines

⚠️ This is a high risk change. Sufficient testing must be done before merging! ⚠️ 

As I run tests on this branch, I will post the results below as comments.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/extension.py
+++ b/datadog_lambda/extension.py
@@ -1,16 +1,6 @@
 import logging
 from os import path
 
-try:
-    # only available in python 3
-    # not an issue since the extension is not compatible with python 2.x runtime
-    # https://docs.aws.amazon.com/lambda/latest/dg/using-extensions.html
-    import urllib.request
-except ImportError:
-    # safe since both calls to urllib are protected with try/expect and will return false
-    urllib = None
-
-AGENT_URL = "http://127.0.0.1:8124"
 HELLO_PATH = "/lambda/hello"
 FLUSH_PATH = "/lambda/flush"
 EXTENSION_PATH = "/opt/extensions/datadog-agent"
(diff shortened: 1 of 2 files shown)


purple4reina:
I created two apps, one using v85 of the prod layer and another using a layer built from this branch. I then curled the two apps on a loop:

```
while true ; do curl [URL1] ; curl [URL2] ; done
```

I let this run for several hours. Here are stats from a 4 hour window of this run:

Screenshot 2023-12-14 at 10 06 50 AM

purple4reina:
I created two apps, one using v85 of the prod layer and another using a layer built from this branch. I then curled the two apps on a loop with a sleep in between:

while true ; do curl [URL1] ; curl [URL2] ; sleep 120 ; done

I let this run for about 30 minutes. Here are stats:

Screenshot 2023-12-14 at 10 16 26 AM

As you can see, the difference is not as large as it was for the busy function. Since there are no errors, I will say that this PR neither hurts nor helps apps like these.

purple4reina:
I created two apps, one using v85 of the prod layer and another using a layer built from this branch. The functions will timeout during init on the first try, but succeed on the second.

```python
import os
import time

filename = '/tmp/testfile'
if not os.path.exists(filename):
    # cause timeout on first invocation only
    with open(filename, 'w') as f:
        f.write('hello')
    time.sleep(600)

def handler(event, context):
    return 'ok'
```

I then curled the two apps on a loop, changing the max memory in between in order to force a cold start.

Screenshot 2023-12-14 at 10 50 04 AM

In neither the prod nor testing layer versions was I able to receive the `aws.lambda.enhanced.init_duration` metric. Therefore, I am unable to state the overhead impact on this scenario.

On the other hand, the function still performs as expected when using the new layer.

purple4reina:
Here is a test with functions that are forced to cold start on each invocation. We can see that there is no significant change in overhead on cold start.

Screenshot 2023-12-14 at 11 27 35 AM

purple4reina:
Closing in favor of https://github.com/DataDog/datadog-lambda-python/pull/406. The `flush` endpoint on the extension does nothing, therefore we shouldn't be calling it in the first place.

Issue #417, opened 2024-02-14
Setting DD_LOG_LEVEL to TRACE crashes python function


MatejBalantic:
## Expected Behavior

No crash when setting environment variable `DD_LOG_LEVEL` to `TRACE`

## Actual Behavior

Crash

## Steps to Reproduce the Problem

I am using DataDog lambda extension. 
I've set the environment variable `DD_LOG_LEVEL` to `TRACE` as one of the allowed values [according to the docs](https://docs.datadoghq.com/agent/troubleshooting/debug_mode/?tab=agentv6v7).

After deployment, lambda crashed in runtime with the following error:
```
[ERROR] ValueError: Unknown level: 'Level TRACE'
```

It seems that the [python handler](https://github.com/DataDog/datadog-lambda-python/blob/b9f8c0fd4ed9359415a3221efb993ea4bf978aba/datadog_lambda/__init__.py#L17) attempts to set python's logger level to mirror `DD_LOG_LEVEL`, but `TRACE` isn't one of allowed levels.

## Specifications

  - Datadog Lambda Layer version: 52
  - Python version: 3.11
  
## Stacktrace

```
  File "/var/lang/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/var/task/datadog_lambda/__init__.py", line 17, in <module>
    logger.setLevel(logging.getLevelName(os.environ.get("DD_LOG_LEVEL", "INFO").upper()))
  File "/var/lang/lib/python3.11/logging/__init__.py", line 1464, in setLevel
    self.level = _checkLevel(level)
[...]

PR #418, opened 2024-02-14
Prevent function crash when setting unknown log level.

purple4reina:
### What does this PR do?

Prevents function crash when setting `DD_LOG_LEVEL` to an unknown value. Also increases the number of available default log levels.

### Motivation

https://github.com/DataDog/datadog-lambda-python/issues/417

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/logger.py
+++ b/datadog_lambda/logger.py
@@ -1,0 +1,27 @@
+import logging
+import os
+
+try:
+    _level_mappping = logging.getLevelNamesMapping()
+except AttributeError:
+    # python 3.8
+    _level_mappping = {name: num for num, name in logging._levelToName.items()}
+# https://docs.datadoghq.com/agent/troubleshooting/debug_mode/?tab=agentv6v7#agent-log-level
+_level_mappping.update(
+    {
+        "TRACE": 5,
+        "WARN": logging.WARNING,
+        "OFF": 100,
+    }
+)
+
+
+def initialize_logging(name):
+    logger = logging.getLogger(name)
+    str_level = (os.environ.get("DD_LOG_LEVEL") or "INFO").upper()
+    level = _level_mappping.get(str_level)
+    if level is None:
+        logger.setLevel(logging.INFO)
+        logger.warning("Invalid log level: %s Defaulting to INFO", str_level)
+    else:
+        logger.setLevel(level)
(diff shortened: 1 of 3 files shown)


datadog_lambda/logger.py
```diff
@@ -0,0 +1,22 @@
...
+_level_mappping.update({
+    # https://docs.datadoghq.com/agent/troubleshooting/debug_mode/?tab=agentv6v7#agent-log-level
+    "TRACE": 5,
+    "WARN": logging.WARNING,
+    "OFF": 100,
+})
+
+def initialize_logging(name):
+    logger = logging.getLogger(name)
+    str_level = os.environ.get("DD_LOG_LEVEL", "INFO").upper() or "INFO"
```
duncanista, line 14: Is the `or "INFO"` ever gonna reach? Since you are already setting the default value of the env var to `INFO`?

purple4reina: It's to cover the `DD_LOG_LEVEL=""` case. I figured this was easier than adding an `if str_level == ""` later on.

purple4reina: Oh, I guess I can remove that first `"INFO"`, that's a good idea.

duncanista: Mmmm, I think the `DD_LOG_LEVEL=""` should be covered by it. We could add it to the tests too, just to confirm.

That also is reached when no env var is set AFAIK.

tests/test_logger.py
```diff
@@ -0,0 +1,43 @@
+import io
+import logging
+import pytest
+
+from datadog_lambda.logger import initialize_logging
+
+_test_initialize_logging = (
```
duncanista, line 7: great parameterizing!

duncanista (approved):
LGTM!

tests/test_logger.py
purple4reina: in python, it's spelled `parametrizing`, whoever wrote it forgot the second `e` 😆

duncanista: I wonder which one is the english version and which is the american one lol

purple4reina: According to wikipedia I guess both spells are correct? And that they're all british? https://en.wiktionary.org/wiki/parametrize

PR #440, opened 2024-03-08
feat: Enable ca-west-1 and il-central-1

### LINKED ISSUE #420: [Feature Request] Add Support for ca-west-1 region
## Expected Behavior
I should be able to use lambda runtime layer in ca-west1 region of AWS

## Actual Behavior
```
╷
│ Error: creating Lambda Function (xxxxxx): operation error Lambda: CreateFunction, https response error StatusCode: 403, RequestID: c0649569-c7d1-44e5-900b-d1b925b6af02, api error AccessDeniedException: User: arn:aws:sts::xxxxx is not authorized to perform: lambda:GetLayerVersion on resource: arn:aws:lambda:ca-west-1:464622532012:layer:Datadog-Python39:78 because no resource-based policy allows the lambda:GetLayerVersion action
```

## Steps to Reproduce the Problem

  1. Attempt to use Datadog-Python39:88  on lambda
  2. Apply Terraform

## Specifications

  - Datadog Lambda Layer version: 88
  - Python version: 3.9
  
## Stacktrace
  
  ```
An error occurred (AccessDeniedException) when calling the GetLayerVersion operation: User: arn:aws:sts::xxxxxx is not authorized to perform: lambda:GetLayerVersion on resource: arn:aws:lambda:ca-west-1:464622532012:layer:Datadog-Python39:88 because no resource-based policy allows the lambda:GetLayerVersion action
  ```

astuyve:
### What does this PR do?
This will cause our CI to publish 8 layers up to 90 versions in two new AWS regions.

### Motivation
Fixes #420 

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/ci/datasources/regions.yaml
+++ b/ci/datasources/regions.yaml
@@ -15,7 +15,7 @@ regions:
   - code: "ap-northeast-2"
   - code: "ap-northeast-3"
   - code: "ca-central-1"
-  - code: "ca-west-1" we don't support it
+  - code: "ca-west-1"
   - code: "eu-central-1"
   - code: "eu-central-2"
   - code: "eu-north-1"
@@ -24,7 +24,7 @@ regions:
   - code: "eu-west-3"
   - code: "eu-south-1"
   - code: "eu-south-2"
-  - code: "il-central-1" we don't support it
+  - code: "il-central-1"
   - code: "me-south-1"
   - code: "me-central-1"
   - code: "sa-east-1"
--- a/ci/datasources/regions.yaml
+++ b/ci/datasources/regions.yaml
@@ -15,7 +15,7 @@ regions:
   - code: "ap-northeast-2"
   - code: "ap-northeast-3"
   - code: "ca-central-1"
-#  - code: "ca-west-1" we don't support it
+  - code: "ca-west-1" we don't support it
   - code: "eu-central-1"
   - code: "eu-central-2"
   - code: "eu-north-1"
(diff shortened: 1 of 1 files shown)


joeyzhao2018 (commented):
I actually think it's safer to just manually do the catch up. And then we enable the new regions... How confident we are about CI doing catch up deployments?

astuyve:
> How confident we are about CI doing catch up deployments?
Not confident.

But that's okay, this is how we'll find out.

This release includes no code changes and only handles these two regions so the downside risk is low. The risk to other regions is low.

If needed we can re-execute the build and re-run the publish scripts. Let's consider the failure modes.

1. (best case) - temporary credentials time out and we need to re-run the individual step for each new region and each runtime/architecture combo. This touches no other regions and allows us to re-execute. This should be pretty painless.
2. Hypothetically let's say the artifact times out of s3 (it's a 24hr window). Then we'd have to re-build it. That's okay too because the jobs for other regions are re-entrant. We do not `version + 1` for production releases, so those other matrix jobs will succeed and be no-ops
3. Let's say this all falls apart and we mismatch versions across regions. We can still version + 2 manually for all regions with no downside risk, as this is effectively equivalent to a regular release. Versions behind would just get + 2 versions (which is our normal catch-up behavior), and the versions which had an extra release will simply +1.

Ultimately we need to get to a place where this happens automatically otherwise we have downstream risks like APIs timing out and mispublishing incompatible arm/x86 versions. I don't believe we have a choice *other* than to try this.

joeyzhao2018:
ok, let's give it a shot. I think I used a wrong word "safer".  I was thinking more like "easier"...

joeyzhao2018 (approved):

astuyve:
I think your wording, reasoning, and caution are correct and well grounded. I also think we're obliged to make this work in CI anyway, so I've de-risked this as much as possible. I will of course watch the deployment very closely.

PR #446, opened 2024-03-12
Properly parent lazily loaded module imports.

purple4reina:
### What does this PR do?

Properly parent `aws.lambda.import` spans for packages that are imported during function execution. Currently app spans are children of the cold start span, which is incorrect.

Screenshot 2024-03-06 at 1 01 36 PM

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/cold_start.py
+++ b/datadog_lambda/cold_start.py
@@ -8,6 +8,7 @@ logger = logging.getLogger(__name__)
 _cold_start = True
 _proactive_initialization = False
 _lambda_container_initialized = False
+_tracer = None
 
 
 def set_cold_start(init_timestamp_ns):
@@ -18,6 +19,7 @@ def set_cold_start(init_timestamp_ns):
     global _cold_start
     global _lambda_container_initialized
     global _proactive_initialization
+    global _tracer
     if not _lambda_container_initialized:
         now = time.time_ns()
         if (now - init_timestamp_ns) // 1_000_000_000 > 10:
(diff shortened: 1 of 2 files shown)


datadog_lambda/cold_start.py
```diff
@@ -29,6 +31,7 @@ def set_cold_start(init_timestamp_ns):
         _cold_start = False
         _proactive_initialization = False
     _lambda_container_initialized = True
+    from ddtrace import tracer as _tracer
```
duncanista, line 34: Didn't know that you could just assign a value by doing an `as` wow

datadog_lambda/cold_start.py
```diff
@@ -70,10 +76,8 @@ def __init__(self, module_name, full_file_path, start_time_ns, end_time_ns=None)
 
 
 def reset_node_stacks():
-    global root_nodes
-    root_nodes = []
-    global import_stack
-    import_stack = []
+    root_nodes.clear()
```
duncanista, line 79: why are we removing the `globals`?

purple4reina: Good question. Replacing the lists with new ones was causing my test to fail because somewhere we're holding onto an old reference to these objects. The correct solution is to just remove all the stuff from inside the list.

duncanista: Yet, this would create an O(n) complexity instead of O(1) of reassigning right? Although I guess we'd have the remaining O(n) space in the wild while reassigning. I tried researching which was the best solution for this, but theres like 5 ways of doing it. So I'll take this one.

joeyzhao2018 (approved):
Nice Work!

datadog_lambda/cold_start.py
purple4reina: This method does not do an allocation where as the previous did.

PR #453, opened 2024-03-27
[SVLS-4647] Fix lambda authorizer trace context injection issue

joeyzhao2018:
### What does this PR do?

There's a regression introduced about authorizer span due to the hardcoded headers values broken by the changes happened in ddtrace.
Specifically, the value of trace id has been directly fetched from the span_context like this `injected_headers[TraceHeader.TRACE_ID] = str(span_context.trace_id)` and that `span_context.trace_id` is now changed into a big int and causing problem when the downstream main lambda tries to extract it
Screenshot 2024-03-26 at 11 43 53 AM
The fix is to use the official propagtor from ddtrace to inject the context.

### Testing Guidelines

Updated the unittest to use the `propagator` too in order to future proof such kind of issue introduced by dependency updates.
Tested in a real lambda case and see it back to normal like this:
Screenshot 2024-03-27 at 1 15 22 PM

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -22,7 +22,6 @@ from datadog_lambda.constants import (
     TraceContextSource,
     XraySubsegment,
     Headers,
-    TraceHeader,
 )
 from datadog_lambda.metric import (
     flush_stats,
@@ -44,6 +43,7 @@ from datadog_lambda.tracing import (
     InferredSpanInfo,
     is_authorizer_response,
     tracer,
+    propagator,
 )
 from datadog_lambda.trigger import (
     extract_trigger_tags,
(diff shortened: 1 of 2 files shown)


datadog_lambda/wrapper.py
```diff
@@ -254,13 +254,7 @@ def _inject_authorizer_span_headers(self, request_id):
...
         source_span = self.inferred_span if self.inferred_span else self.span
         span_context = source_span.context
-        injected_headers[TraceHeader.TRACE_ID] = str(span_context.trace_id)
-        injected_headers[TraceHeader.PARENT_ID] = str(span_context.span_id)
-        sampling_priority = span_context.sampling_priority
-        if sampling_priority is not None:
-            injected_headers[TraceHeader.SAMPLING_PRIORITY] = str(
-                span_context.sampling_priority
-            )
+        propagator.inject(span_context, injected_headers)
```
duncanista, line 257: Great catch, I wonder if this methods should eventually be refactored into a tracing service and not the wrapper 😕

duncanista: the `propagator.inject` ends up injecting into `injected_headers`, correct?

duncanista (approved):
LGTM – left a non-blocking question.

datadog_lambda/wrapper.py
joeyzhao2018: Good question! 

Having the same thought, I went ahead trying to refactor it. 
Then I found it's already in a helper method called `_inject_authorizer_span_headers`.
Then it felt that the `injection` is more relevant to the `wrapper` because the wrapper now has the `inferred_span` or `span` data and it also should be responsible for the final `response` data. And then at this point, the `injection` became more like `prepare_response` and the only tracer related part is already encapsulated by the `propagator.inject` helper.
Hope this makes sense.

PR #454, opened 2024-03-27
[SVLS-4298] Java upstream sqs trace context propagation

joeyzhao2018:
### What does this PR do?

This is python's version of the same logic done in the NodeJS layer PR https://github.com/DataDog/datadog-lambda-js/pull/511

### Motivation

JAVA lambda would put trace context information in the `AWSTracerHeader` inside the `message.attributes`
This PR extract that so that  JAVA Lambda => sqs => python lambda would show as one connected trace.

### Additional Notes

I refrained from refactoring too much of the code logic in this PR. But I plan to do some refactoring in a following PR. This is because the refactoring would focus more on the general structure of the extractors. Therefore not very relevant to the java upstream case here.

### Testing Guidelines

Added unittest case using a java upstream example payload.
Manually tested in a real lambda setup:
Screenshot 2024-03-27 at 2 02 08 PM

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -291,16 +291,11 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                     logger.debug(
                         "Found dd-trace injected trace context from AWSTraceHeader"
                     )
-                    dd_data = {
-                        TraceHeader.TRACE_ID: str(
-                            int(trace_id_parts[2][8:], 16)
-                        ),  # remove padding and convert the hex str to a decimal str
-                        TraceHeader.PARENT_ID: str(
-                            int(x_ray_context["parent_id"], 16)
-                        ),  # convert the hex str to a decimal str
-                        TraceHeader.SAMPLING_PRIORITY: x_ray_context["sampled"],
-                    }
-                    return propagator.extract(dd_data)
+                    return Context(
+                        trace_id=int(trace_id_parts[2][8:], 16),
+                        span_id=int(int(x_ray_context["parent_id"], 16)),
+                        sampling_priority=float(x_ray_context["sampled"]),
+                    )
         return extract_context_from_lambda_context(lambda_context)
     except Exception as e:
         logger.debug("The trace extractor returned with error %s", e)
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -69,6 +69,8 @@ if dd_tracing_enabled:
 
 propagator = HTTPPropagator()
 
+DD_TRACE_JAVA_TRACE_ID_PADDING = "00000000"
+
 
 def _convert_xray_trace_id(xray_trace_id):
     """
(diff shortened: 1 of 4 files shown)


datadog_lambda/tracing.py
```diff
@@ -248,28 +249,55 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
-            return extract_context_from_lambda_context(lambda_context)
-        dd_data = json.loads(dd_json_data)
-        return propagator.extract(dd_data)
+            # Handle case where trace context is injected into attributes.AWSTraceHeader
+            # example: Root=1-654321ab-000000001234567890abcdef;Parent=0123456789abcdef;Sampled=1
+            x_ray_header = first_record.get("attributes", {}).get("AWSTraceHeader")
+            if x_ray_header:
+                x_ray_context = parse_xray_header(x_ray_header)
+                trace_id_parts = x_ray_context.get("trace_id", "").split("-")
+                if len(trace_id_parts) > 2 and trace_id_parts[2].startswith("00000000"):
```
duncanista, line 284: turn magic 0s into constant like in JS?

duncanista (approved):
LGTM – left a non-blocking nit. 

Looking forward for the refactor 🙏🏽  lmk if you need help there!

datadog_lambda/tracing.py
joeyzhao2018: i did it but then reverted it somehow... cannot remember the reason now... so I'll just use a constant as you mentioned. 👍

datadog_lambda/tracing.py
```diff
@@ -248,28 +249,55 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                    dd_data = {
+                        TraceHeader.TRACE_ID: str(
+                            int(trace_id_parts[2][8:], 16)
+                        ),  # remove padding and convert the hex str to a decimal str
+                        TraceHeader.PARENT_ID: str(
+                            int(x_ray_context["parent_id"], 16)
+                        ),  # convert the hex str to a decimal str
+                        TraceHeader.SAMPLING_PRIORITY: x_ray_context["sampled"],
+                    }
+                    return propagator.extract(dd_data)
```
purple4reina, line 303: Instead of calling extract, why not just manually create the `Context` object? For example, this code here will not work if the extract style is just `w3c`.

joeyzhao2018: Updated it and tested a testing layer with the java upstream lambda having `DD_TRACE_PROPAGATION_STYLE_INJECT = tracecontext` and the downstream python lambda having `DD_TRACE_PROPAGATION_STYLE_EXTRACT=tracecontext`. 
Screenshot 2024-03-27 at 2 52 58 PM

purple4reina (approved):

PR #456, opened 2024-03-27
Parametrized trace context extraction tests.

purple4reina:
### What does this PR do?

Parameterize some of the trace extraction tests.

### Motivation

Clearer. Easier. Prettier.

### Additional Notes

There are still more trace extraction tests in there, I am thinking to move those others into this single parameterized test soon.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -96,6 +96,58 @@ def with_trace_propagation_style(style):
     return _wrapper
 
 
+_test_extract_dd_trace_context = (
+        ("api-gateway", Context(trace_id=12345, span_id=67890, sampling_priority=2)),
+        ("api-gateway-no-apiid", Context(trace_id=12345, span_id=67890, sampling_priority=2)),
+        ("api-gateway-non-proxy", Context(trace_id=12345, span_id=67890, sampling_priority=2)),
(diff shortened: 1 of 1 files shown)


tests/test_tracing.py
```diff
@@ -96,6 +96,58 @@ def _wrap(*args, **kwargs):
     return _wrapper
 
 
+_test_extract_dd_trace_context = (
+        ("api-gateway", Context(trace_id=12345, span_id=67890, sampling_priority=2)),
```
duncanista, line 100: Question, is there any difference if we just use the same Context object for all of them? I.e. data is the same, as opposed to changing the trace_id, span_id and sampling_priority all the time? 

I see some of them have different sampling priority, but I don't understand why they have different trace_id/span_id, is it important in those cases?

purple4reina: I suspect the reason why the trace/span ids are different is because these values are pulled from the event json itself from the text fixture file. Bringing these into alignment would mean changing those fixtures and could have impacts on other tests as well.

joeyzhao2018 (approved):

tests/test_tracing.py
duncanista: Thanks – makes sense, appreciate it

PR #458, opened 2024-03-29
Replace string formatting with f strings.

purple4reina:
### What does this PR do?

Replaces string formatting `"my {}".format("string")` with f strings `f"my {'string'}"`.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/cold_start.py
+++ b/datadog_lambda/cold_start.py
@@ -50,14 +50,14 @@ def is_new_sandbox():
 
 def get_cold_start_tag():
     """Returns the cold start tag to be used in metrics"""
-    return "cold_start:true" if is_cold_start() else "cold_start:false"
+    return "cold_start:true" if _cold_start else "cold_start:false"
 
 
 def get_proactive_init_tag():
     """Returns the proactive init tag to be used in metrics"""
     return (
         "proactive_initialization:true"
-        if is_proactive_init()
+        if _proactive_initialization
         else "proactive_initialization:false"
     )
 
--- a/datadog_lambda/cold_start.py
+++ b/datadog_lambda/cold_start.py
@@ -50,12 +50,16 @@ def is_new_sandbox():
 
 def get_cold_start_tag():
     """Returns the cold start tag to be used in metrics"""
-    return "cold_start:{}".format(str(is_cold_start()).lower())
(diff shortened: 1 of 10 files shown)


datadog_lambda/cold_start.py
```diff
@@ -50,12 +50,16 @@ def is_new_sandbox():
 
 def get_cold_start_tag():
     """Returns the cold start tag to be used in metrics"""
-    return "cold_start:{}".format(str(is_cold_start()).lower())
+    return "cold_start:true" if is_cold_start() else "cold_start:false"
```
duncanista, line 53: Can't we just?
```suggestion
    return f"cold_start:{is_cold_start()}"
```

datadog_lambda/cold_start.py
```diff
@@ -50,12 +50,16 @@ def is_new_sandbox():
...
 
 
 def get_proactive_init_tag():
     """Returns the proactive init tag to be used in metrics"""
-    return "proactive_initialization:{}".format(str(is_proactive_init()).lower())
+    return (
+        "proactive_initialization:true"
+        if is_proactive_init()
```
duncanista, line 60: same as the cold start one?

astuyve (approved):

datadog_lambda/cold_start.py
purple4reina: Sure, but why do all that work when we already know what the tag will be?

Using python's `timeit` module:

```bash
$ python -m timeit -s 'from datadog_lambda.cold_start import is_cold_start' '"cold_start:true" if is_cold_start() else "cold_start:false"'
10000000 loops, best of 5: 25.2 nsec per loop
$ python -m timeit -s 'from datadog_lambda.cold_start import is_cold_start' 'f"cold_start{is_cold_start()}"'
5000000 loops, best of 5: 86.4 nsec per loop

Comment (purple4reina) [reply datadog_lambda/cold_start.py:53]: Oh, I found an even faster way!!

```python
return "cold_start:true" if _cold_start else "cold_start:false"
```

Cuts it down to `24.7 nsec per loop` from `40.9 nsec per loop`!

duncanista: sounds good, thanks for the explanation

PR #462, opened 2024-04-01
Do not recalculate enhanced metric setting each time.

purple4reina:
### What does this PR do?

Sets the value for enhanced metrics as a global instead of recalculating it each time.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/extension.py
+++ b/datadog_lambda/extension.py
@@ -1,5 +1,5 @@
 import logging
-from os import path
+import os
 
 AGENT_URL = "http://127.0.0.1:8124"
 FLUSH_PATH = "/lambda/flush"
@@ -9,9 +9,7 @@ logger = logging.getLogger(__name__)
 
 
 def is_extension_present():
-    if path.exists(EXTENSION_PATH):
-        return True
-    return False
+    return os.path.exists(EXTENSION_PATH)
 
 
 def flush_extension():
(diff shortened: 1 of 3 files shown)


datadog_lambda/extension.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from os import path
+import os
```
duncanista, line 2: why importing the whole`os` instead of `path` from `os`?

datadog_lambda/metric.py
```diff
@@ -32,6 +32,10 @@
     flush_in_thread = os.environ.get("DD_FLUSH_IN_THREAD", "").lower() == "true"
     lambda_stats = ThreadStatsWriter(flush_in_thread)
 
+enhanced_metrics_enabled = (
```
duncanista, line 35: why a tuple?

datadog_lambda/extension.py
purple4reina: habit. I'm just used to seeing `os.path.exists` not `path.exists`.

datadog_lambda/metric.py
purple4reina: It's not a tuple. Tuples have a comma in them. Without the comma it's just a way to put something on multiple lines.

duncanista: Oop, tysm :til:

duncanista (approved):

PR #463, opened 2024-04-02
Reduce allocations during event parsing.

purple4reina:
### What does this PR do?

Reduces allocations in all functions in the `datadog_lambda/tracing.py` file.

### Motivation

It's faster.

### Additional Notes

Screenshot 2024-04-02 at 3 03 39 PM

Before and after comparison. Note this only compares the parsing of lambda function urls. Therefore, other event types could see more/less performance improvements.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -253,19 +253,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
             # SQS uses dataType and binaryValue/stringValue
             # SNS uses Type and Value
             dd_json_data = None
-            dd_json_data_type = dd_payload.get("Type")
-            if dd_json_data_type is None:
-                dd_json_data_type = dd_payload.get("dataType")
+            dd_json_data_type = dd_payload.get("Type") or dd_payload.get("dataType")
             if dd_json_data_type == "Binary":
-                dd_json_data = dd_payload.get("binaryValue")
-                if dd_json_data is None:
-                    dd_json_data = dd_payload.get("Value")
+                dd_json_data = dd_payload.get("binaryValue") or dd_payload.get("Value")
                 if dd_json_data:
                     dd_json_data = base64.b64decode(dd_json_data)
             elif dd_json_data_type == "String":
-                dd_json_data = dd_payload.get("stringValue")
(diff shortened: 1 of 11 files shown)


datadog_lambda/tracing.py
```diff
@@ -66,6 +66,8 @@
 
         telemetry_writer.enable()
 
+is_lambda_context = os.environ.get(XrayDaemon.FUNCTION_NAME_HEADER_NAME) != ""
```
purple4reina, line 69: This logic used to be in a function. But since the function name header won't change between invocations, we can parse it just once at the global level.

joeyzhao2018 (approved):
LGTM. Moreover, I actually overlooked allocations such as `get("key", _some_default)`. We should probably summarize these things as a guide and add it to the PR template.

datadog_lambda/tracing.py
```diff
@@ -235,37 +235,37 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                if dd_json_data:
+                    dd_json_data = base64.b64decode(dd_json_data)
             elif dd_json_data_type == "String":
-                dd_json_data = dd_payload.get(
-                    "stringValue",
-                    dd_payload.get("Value", r"{}"),
-                )
+                dd_json_data = dd_payload.get("stringValue")
+                if dd_json_data is None:
+                    dd_json_data = dd_payload.get("Value")
```
duncanista, line 268: Although I like that we are reducing allocations, this is really hard to read :cry: maybe in the future when we split the files, this would make more sense

duncanista (approved):

datadog_lambda/tracing.py
purple4reina: Yes agreed. This is difficult to read. I'll take a stab at improving it a bit.

duncanista: Another PR is fine, I guess when this is separated, this can be much easier! Not a blocking comment

purple4reina: @duncanista Okay, I made some updates. Let me know if this is any better for you.

PR #467, opened 2024-04-09
Use single socket instance for all xray api calls.

purple4reina:
### What does this PR do?

+ Use single socket connection for all interactions with the xray api
+ Memoize xray api host/port identification
+ Ensure socket creation is done as lazily as possible

### Motivation

The need for speed.

### Additional Notes

Previously, we were creating a new socket connection and closing the connection for each xray api call. When xray is enabled, and the lambda has extracted trace context from an upstream event, the xray api is called twice.

When an xray api is called, this change decreases runtime duration by somewhere between 1-5%.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/xray.py
+++ b/datadog_lambda/xray.py
@@ -10,32 +10,52 @@ from datadog_lambda.constants import XrayDaemon, XraySubsegment, TraceContextSou
 logger = logging.getLogger(__name__)
 
 
-def get_xray_host_port(address):
-    if address == "":
-        logger.debug("X-Ray daemon env var not set, not sending sub-segment")
-        return None
-    parts = address.split(":")
-    if len(parts) <= 1:
-        logger.debug("X-Ray daemon env var not set, not sending sub-segment")
-        return None
-    port = int(parts[1])
-    host = parts[0]
-    return (host, port)
+class Socket(object):
+    def __init__(self):
+        self.sock = None
+
+    @property
+    def host_port_tuple(self):
+        if not hasattr(self, "_host_port_tuple"):
+            self._host_port_tuple = self._get_xray_host_port(
+                os.environ.get(XrayDaemon.XRAY_DAEMON_ADDRESS, "")
+            )
+        return self._host_port_tuple
+
+    def send(self, payload):
+        if not self.sock:
+            self._connect()
+        try:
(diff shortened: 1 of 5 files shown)


datadog_lambda/xray.py
```diff
@@ -10,32 +10,52 @@
...
+
+    def send(self, payload):
+        if not self.sock:
+            self._connect()
+        try:
+            self.sock.send(payload.encode("utf-8"))
+        except Exception as e_send:
+            logger.error("Error occurred submitting to xray daemon: %s", e_send)
+
+    def reset(self):
```
purple4reina, line 33: `reset` is used for tests.

datadog_lambda/xray.py
```diff
@@ -10,32 +10,52 @@
...
+    def __init__(self):
+        self.sock = None
+
+    @property
+    def host_port_tuple(self):
+        if not hasattr(self, "_host_port_tuple"):
+            self._host_port_tuple = self._get_xray_host_port(
+                os.environ.get(XrayDaemon.XRAY_DAEMON_ADDRESS, "")
+            )
+        return self._host_port_tuple
```
purple4reina, line 23: This value is now memoized so we only need to parse the env var once.

tests/test_wrapper.py
```diff
@@ -623,7 +626,7 @@ def handler(event, context):
         self.assertEqual(result.span_id, aws_lambda_span.span_id)
         self.assertEqual(result.sampling_priority, 1)
         mock_send.assert_called_once()
-        (_, raw_payload), _ = mock_send.call_args
+        (raw_payload,), _ = mock_send.call_args
```
purple4reina, line 629: Signature of the `xray.send` method changed.

datadog_lambda/xray.py
joeyzhao2018: nitpick: can we move it to tests files?

joeyzhao2018 (approved):
LGTM

datadog_lambda/xray.py
purple4reina: Yes, good call. Done.

PR #474, opened 2024-04-17
Do not include instrumentation for test runners.

purple4reina:
### What does this PR do?

Removes the following instrumentation libraries from ddtrace:
+ `pytest`
+ `pytest_bdd`
+ `pytest_benchmark`
+ `unittest`
+ `coverage`

### Motivation

Reduces size of the layer.

Before
```
Layer file .layers/datadog_lambda_py-amd64-3.9.zip has zipped size 3480 kb
Layer file .layers/datadog_lambda_py-amd64-3.9.zip has unzipped size 10485 kb
```

After
```
Layer file .layers/datadog_lambda_py-amd64-3.9.zip has zipped size 3452 kb
Layer file .layers/datadog_lambda_py-amd64-3.9.zip has unzipped size 10393 kb
```

### Additional Notes

These packages are only necessary if customers are running their python tests in lambda and have enabled the testing ci tools. If they are doing this and still want these instrumentations, then they can always vendor their own version of ddtrace.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

https://datadoghq.atlassian.net/browse/SVLS-4701

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,9 @@ RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/li
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/ddup/_ddup.*.so
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/stack_v2/_stack_v2.*.so
 
+RUN rm -rf ./python/lib/$runtime/site-packages/ddtrace/contrib/pytest*
+RUN rm -rf ./python/lib/$runtime/site-packages/ddtrace/contrib/unittest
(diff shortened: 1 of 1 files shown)


duncanista:
Although I like the idea – wouldn't this be technically a breaking change?

purple4reina:
> Although I like the idea – wouldn't this be technically a breaking change?

Yeah. It would be a breaking change. We'd need to add this to the release notes to make sure customers are aware of it.

Given that the size decrease isn't huge (about 30mb compressed) I think it's worth considering if we should do this at all.

joeyzhao2018:
Are those the only things we can strip from `ddtrace.contrib`? I'm thinking if we are doing this (breaking change) let's get them all! WDYT?

purple4reina:
Closing this for now since the size impact isn't that noticeable. We can revisit later if we need to claw 28 kb back.

Issue #210, opened 2022-03-08
When a line of code raises outside of the handler function, then datadog is not able to detect the error.


nalepae:
## Expected Behavior
When a line of code raises **outside** of the handler function, then `datadog` should be able to detect the error.

## Actual Behavior
When a line of code raises **outside** of the handler function, then `datadog` is not able to detect the error.
==> If an monitor (attached to a Slack alert) is set up when an exception is raised (and not catched) on this lambda, then the corresponding monitor is not triggered and no Slack message is sent.

## Steps to Reproduce the Problem
Define the following Lambda function:

```Python
import json

# The following line will raise on purpose
0/0

def lambda_handler(event, context):
    return {
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!')
    }
```

Set 
- The AWS handler to `datadog_lambda.handler.handler`, and
- The `DD_LAMBDA_HANDLER` environment variable to `<your_file>.lambda_handler`

Run the lambda.
==> Even if the line `0/0` raises, no trace will be visible in the **Invocation Serverless** part of Datadog.
![image](https://user-images.githubusercontent.com/4943830/157214305-14345b5a-1355-43d3-a07b-3b4a1054b1ac.png)

Note we see invocations on top left chart (3 blue vertical bars), but there is nothing in the center panel (`No traced invocation in the time window`), no way to see the traces, the Python stack trace ...

If we move the `0/0` in the handler, like below:
```Python
import json

def lambda_handler(event, context):
    # The following line will raise on purpose
    0/0
    return {
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!')
    }
```

then Datadog behaves correctly (visible Traces, Monitor, Slack Message ...)

## Specifications

  - Datadog Lambda Layer version: 
![image](https://user-images.githubusercontent.com/4943830/157215837-b23d0882-e408-4068-af26-38354cb79cd8.png)
 
  - Python version: 3.8
  
## Additional information
I understand we specify the handler to Datadog, and thus cannot be aware of things running out of the handler, but as indicated in [AWS best practices](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html), there is benefits to run some code out of the handler. If this code fails, it is very important that the developer team is notified.

[...]

PR #475, opened 2024-04-17
Detect exceptions raised outside of the handler

TalUsvyatsky:
### What does this PR do?

This PR adds telemetry for exceptions that occur outside of the handler function. 

When an exception occurs outside of the handler function:
1. `aws.lambda.enhanced.errors` is emitted
2. A trace with an error span is created

### Motivation

- https://datadoghq.atlassian.net/browse/SVLS-4666
- Users should have visibility into this category of errors.
- #210 

### Testing Guidelines
1. Created a handler file like this:
```python
  import json
 
  0/0 # Purposely raise an exception

  def hello(event, context):
      body = {
          "message": "some message",
          "input": event,
      }
      return {"statusCode": 200, "body": json.dumps(body)}

```
2. Deployed my function and invoked it via function URL
3. Saw traces with an error span from the call ([example trace](https://ddserverless.datadoghq.com/apm/trace/15758361503635128623?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5267772117214879748&spanViewType=errors&timeHint=1713881382599.7651))
4. Saw `aws.lambda.enhanced.errors` emitted ([graph](https://ddserverless.datadoghq.com/s/61ba5v0dnxcpd8dm/f47-xyc-kw4))
5. Traces appear in serverless detail page invocation table ([link to search for function](https://ddserverless.datadoghq.com/functions?search=functionname%3A%2Atal-dev%2A%20region%3Asa-east-1&cloud=aws&entity_view=lambda_functions&fromUser=false&panel_end=1713817896171&panel_paused=false&panel_start=1713731496171&panel_tab=errorTracking&shouldShowLegend=true&sort=time&spanViewType=logs&view=spans&start=1713731134137&end=1713817534137&paused=false))
6. Issues appear in error tracking and are informative

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

https://datadoghq.atlassian.net/browse/SVLS-4666


datadog_lambda/handler.py
```diff
@@ -27,5 +51,11 @@ class HandlerError(Exception):
...
-handler_module = import_module(modified_mod_name)
-handler = datadog_lambda_wrapper(getattr(handler_module, handler_name))
+
+try:
+    start_time_ns = time.time_ns()
+    handler_module = import_module(modified_mod_name)
+    handler_func = getattr(handler_module, handler_name)
+    handler = datadog_lambda_wrapper(handler_func)
+except Exception as e:
+    handler = _ErrorOutsideHandlerDecorator(e, modified_mod_name, start_time_ns)
```
joeyzhao2018, line 61: Just want to confirm that we've checked start_time_ns reflects each invocation correctly?

datadog_lambda/handler.py
```diff
@@ -7,6 +7,9 @@
 from importlib import import_module
 
 import os
+import time
+
+from datadog_lambda.tracing import emit_telemetry_on_exception_outside_of_handler
```
duncanista, line 12: Should we lazy load this in the catch? 

What's the overhead difference if we do that?

datadog_lambda/handler.py
```diff
@@ -15,6 +18,27 @@ class HandlerError(Exception):
     pass
 
 
+class _ErrorOutsideHandlerDecorator(object):
```
duncanista, line 21: Why a decorator and not just calling the function?

I'm trying to understand the reason behind it, but this just looks like a wrapper to the `emit_telemetry_on_exception_outside_of_handler` which then re-raises the error.

tests/test_handler.py
```diff
@@ -0,0 +1,80 @@
+import os
```
duncanista, line 1: Thanks for adding unit tests for this whole module!

datadog_lambda/handler.py
TalUsvyatsky: Good call out, I'll double check this

datadog_lambda/handler.py
TalUsvyatsky: Good question, the purpose of the decorator is to get the request ID into the trace (necessary for the frontend to display traces in the invocation table). If we just call `emit_telemetry_on_exception_outside_of_handler` and re-raise the exception on line 61, we won't have access to the lambda context which holds the request ID

datadog_lambda/handler.py
TalUsvyatsky: Oh yeah, good point. What's the best way to measure the difference?

datadog_lambda/handler.py
duncanista: Ah, interesting, you are right. Thanks for explaining, that makes a lot of sense!

datadog_lambda/handler.py
duncanista: Instrument a function with it, check the overall cold start.

Then lazy load it, build it again and instrument a function with it, check the cold start.

Compare them and see the differences.

Since you won't have the cold start tracer here, you might not be able to see the actual imports. So your best chance is to look for CloudWatch Logs to see the cold start here.

datadog_lambda/handler.py
duncanista: Wait, maybe the naming is not ideal. This is not a decorator, this is a fallback-handler for this scenario. What do you think? Maybe I'm wrong, but we are not decorating any handler here, we are creating a new handler due to the customers code not being able to be imported.

TalUsvyatsky: I agree, updated to `_ErrorFallbackHandler` instead, but open to other naming ideas if you've got something in mind

datadog_lambda/handler.py
```diff
@@ -15,6 +18,27 @@ class HandlerError(Exception):
...
+        self.start_time_ns = start_time_ns
+
+    def __call__(self, event, context, **kwargs):
+        emit_telemetry_on_exception_outside_of_handler(
+            context,
+            self.exception,
+            self.modified_mod_name,
+            self.start_time_ns,
+        )
+        raise self.exception
```
purple4reina, line 39: This `_ErrorOutsideHandlerDecorator` is incredibly clever. Very well done.

I'd recommend we move this to `datadog_lambda.wrapper`. There's already a couple of wrapper classes in there. Then, rename it to something that similarly matches the casing etc of `datadog_lambda_wrapper`.

datadog_lambda/handler.py
```diff
@@ -27,5 +51,11 @@ class HandlerError(Exception):
...
-handler_module = import_module(modified_mod_name)
-handler = datadog_lambda_wrapper(getattr(handler_module, handler_name))
+
+try:
+    start_time_ns = time.time_ns()
+    handler_module = import_module(modified_mod_name)
+    handler_func = getattr(handler_module, handler_name)
+    handler = datadog_lambda_wrapper(handler_func)
+except Exception as e:
+    handler = _ErrorFallbackHandler(e, modified_mod_name, start_time_ns)
```
purple4reina, line 61: I just want to call out that this changes the behavior of the function slightly. Now instead of failing during the init phase, it will fail during the execution phase. Totally not a big deal, but good to point out that customers aren't charged for init duration, only for duration duration. This means whereas before they were not charged for these failures, now they will be.

purple4reina:
FYI, pulling from main should fix the check layer size test failure.

datadog_lambda/wrapper.py
```diff
@@ -382,9 +383,31 @@ def _after(self, event, context):
...
+        raise self.exception
+
+
 def format_err_with_traceback(e):
     tb = traceback.format_exc().replace("\n", "\r")
     return f"Error {e}. Traceback: {tb}"
 
 
 datadog_lambda_wrapper = _LambdaDecorator
+error_fallback_handler = _ErrorFallbackHandler
```
purple4reina, line 413: Not to belabor the point too much, but wrapper/decorator/handler, all words we use to refer to the same thing? That's potentially confusing for our future selves. Honestly, I would prefer we match the existing naming as much as we can, however wrong it may be.

Therefore, this line would become
```py
error_fallback_wrapper = _ErrorFallbackDecorator
```

But, also this is really nitpicky, so it's fine to keep it as is too.

purple4reina (approved):

datadog_lambda/handler.py
TalUsvyatsky: Ah I see, that makes sense. Thanks for pointing that out!

datadog_lambda/wrapper.py
duncanista: This is not a wrapper/decorator, that's why I suggested to not name it as a Decorator!

datadog_lambda/handler.py
TalUsvyatsky: This was a good catch, the start time / duration were being set incorrectly. I update the code to set the span start time based on initial time spent importing the handler + time spent in the call itself.

datadog_lambda/handler.py
TalUsvyatsky: This got refactored a bit, but I tested moving the import of `error_fallback_handler` into the catch. The traces actually do get a cold start span, and the durations look right based on the CloudWatch logs. Based on those spans ([trace from import at the top](https://ddserverless.datadoghq.com/apm/trace/14945367174940455538?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4795421708683935480&spanViewType=metadata&timeHint=1713916422872), [trace from import in the catch](https://ddserverless.datadoghq.com/apm/trace/13634600884944720567?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7025402365249368892&spanViewType=metadata&timeHint=1713967321993)), I'm not seeing a noticeable difference in overhead

datadog_lambda/tracing.py
```diff
@@ -1320,3 +1322,39 @@ def is_async(span: Span) -> bool:
...
+
+def emit_telemetry_on_exception_outside_of_handler(
+    context, exception, resource_name, handler_load_duration
+):
+    """
+    Emit an enhanced error metric and create a span for exceptions occuring outside of the handler
+    """
+    submit_errors_metric(context)
+
+    span = tracer.trace(
```
joeyzhao2018, line 1335: We support an environment variable named `DD_TRACE_ENABLED`. I'm thinking we should probably consider that flag for this feature too? WDYT?

datadog_lambda/wrapper.py
purple4reina: Ah yes, I see what you mean. We are not passing the original handler to this `_ErrorFallbackHandler` and thus it is not a wrapper/decorator. Semantics. No strong opinion from me on what we should do here, I'm fine if we keep it as is.

duncanista (approved):
Amazing work – LGTM.

There's one nit:
https://github.com/DataDog/datadog-lambda-python/blob/21012cb8b98a81b96d329583e0bb2fc05a8a9bf5/datadog_lambda/tracing.py#L1236-L1255

Should we care about some of the tags that can be added from this? i.e. `functionname`, `cold_start` (which should always be true if we are loading), `datadog_lambda_version`, `ddtrace_version`

duncanista (approved):
LGTM – left a nit.

purple4reina (approved):

joeyzhao2018 (approved):
LGTM 👍

PR #476, opened 2024-04-18
Pre-compile all python code and remove original .py files.

purple4reina:
## What does this PR do?

Ships python layer with only pre-built bytecode binaries and none of the original `*.py` files.

## Motivation

For python 3.12, reduces cold start by >24%. Also reduces memory overhead and runtime duration by small but statistically significant amounts.

Screenshot 2024-04-17 at 3 15 40 PM

Similar overhead improvements are seen with python 3.9.

Screenshot 2024-04-17 at 2 55 32 PM

### Impact of prebuilding bytecode binaries

When the python interpreter starts up, it always first looks for any `*.pyc` files and if found, loads those instead of the `*.py` files. If it is loading any `*.py` files, it will first compile those into bytecode and save to disk so that the next call is faster. See [flow chart](https://peps.python.org/pep-3147/#flow-chart):

flow

Note that these `*.pyc` files are not actually "compiled" code, because python is still an interpreted language. They are also not platform dependent, meaning we can build these bytecode files anywhere and expect them to work everywhere.

These bytecode files do not change the running/execution of the python code. The code will still run exactly the same and at the same speed. What changes instead is the time it takes to load the instructions, prior to them being executed.

### Impact of removing `*.py` files

In addition to pre-compiling all `*.pyc` files, this PR also removes all `*.py` files. As can be seen from flow chart and discussion above, this will make no change to the execution of the code.

What is potentially lost is the ability for the interpreter to add proper line numbers and code samples to tracebacks. TODO: confirm this and add example here.

### Impact of compiling at "optimization level" 2

In addition to pre-compiling bytecode files, this PR also builds those files using "optimization level" 2 by including `-o 2` in the commandline args.

From the [python docs](https://docs.python.org/3.11/using/cmdline.html#cmdoption-O)

> -O
> Remove assert statements and any code conditional on the value of [__debug__](https://docs.python.org/3.11/library/constants.html#debug__). Augment the filename for compiled ([bytecode](https://docs.python.org/3.11/glossary.html#term-bytecode)) files by adding .opt-1 before the .pyc extension (see [PEP 488](https://peps.python.org/pep-0488/)). See also [PYTHONOPTIMIZE](https://docs.python.org/3.11/using/cmdline.html#envvar-PYTHONOPTIMIZE).
>
> > Changed in version 3.5: Modify .pyc filenames according to [PEP 488](https://peps.python.org/pep-0488/).
>
> -OO
> Do [-O](https://docs.python.org/3.11/using/cmdline.html#cmdoption-O) and also discard docstrings. Augment the filename for compiled ([bytecode](https://docs.python.org/3.11/glossary.html#term-bytecode)) files by adding .opt-2 before the .pyc extension (see [PEP 488](https://peps.python.org/pep-0488/)).
>
> > Changed in version 3.5: Modify .pyc filenames according to [PEP 488](https://peps.python.org/pep-0488/).

*Assert statements.* On April 18, 2024 I attended the python guild meeting and discussed this change with the team. They said that ddtrace should not require execution of `assert` statements in order to run properly. If there is a failure due to the remove of the `assert` statements, then this is considered a bug and should be filed with them directly. I ran several of the ddtrace unittest suites using `python -OO -m pytest` and found that the same tests failed using this as did when I ran the tests using `python -m pytest`.

*`__debug__`.* The `-o 2` flag also removes all code conditional on `__debug__`. I did a grep of ddtrace and found no instances of this variable in the code base. I also did a grep for it in all repos I currently have checked out on my laptop and did not find any instances which would affect us.

*Docstrings.* It is safe to remove the docstrings. The only time the customer would care to use them is in a python shell where they can run `help(my_func)` which will then print the docstring to the screen. This is clearly not necessary for the running of any lambda function. Removing them also reduces the size of the layer binary by a noticeable amount.

### Impact of compiling with env var `PYTHONNODEBUGRANGES=1`

This env var is only available in python 3.11 and newer. See [python docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONNODEBUGRANGES):

> PYTHONNODEBUGRANGES
> If this variable is set, it disables the inclusion of the tables mapping extra location information (end line, start column offset and end column offset) to every instruction in code objects. This is useful when smaller code objects and pyc files are desired as well as suppressing the extra visual location indicators when the interpreter displays tracebacks.
>
> > New in version 3.11.

Adding this env var leads to a noticeable reduction in cold start time (approx 0.5-1%) for python 3.12. I believe what is lost by including this is the column information displayed in any traceback. As these are not required, the trade off here (debug info for cold start improvement) I believe is worth it.

## Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Check all that apply

https://datadoghq.atlassian.net/browse/SVLS-4713


purple4reina:
## Errors

We can see what is lost because we've removed the `*.py` files. Line numbers are still present in the tracebacks. What is lost is the line of code itself. I would argue that these are unnecessary and losing them is worth the cost.

Chances are errors are not going to be caused by our code and thus would just be a distraction to helping the user find the real origin of their error. Also, if they really want to see the line of code, it's all open sourced.

### Raw
[Before](https://ddserverless.datadoghq.com/apm/error-tracking?query=%40resource_names%3Arey-python-lambda-dev-before&fromUser=false&issueId=171ab804-eae1-11ee-9013-da7ad0900002&refresh_mode=sliding&view=spans&from_ts=1713469797633&to_ts=1713473397633&live=true)
Screenshot 2024-04-18 at 1 50 24 PM

[After](https://ddserverless.datadoghq.com/apm/error-tracking?query=%40resource_names%3Arey-python-lambda-dev-after&fromUser=false&issueId=171ab804-eae1-11ee-9013-da7ad0900002&refresh_mode=sliding&view=spans&from_ts=1713469797633&to_ts=1713473397633&live=true)
Screenshot 2024-04-18 at 1 51 45 PM

### Pretty
[Before](https://ddserverless.datadoghq.com/apm/error-tracking?query=%40resource_names%3Arey-python-lambda-dev-before&fromUser=false&issueId=171ab804-eae1-11ee-9013-da7ad0900002&refresh_mode=sliding&view=spans&from_ts=1713469797633&to_ts=1713473397633&live=true)
Screenshot 2024-04-18 at 1 50 30 PM

[After](https://ddserverless.datadoghq.com/apm/error-tracking?query=%40resource_names%3Arey-python-lambda-dev-after&fromUser=false&issueId=171ab804-eae1-11ee-9013-da7ad0900002&refresh_mode=sliding&view=spans&from_ts=1713469797633&to_ts=1713473397633&live=true)
Screenshot 2024-04-18 at 1 51 51 PM

Dockerfile
```diff
@@ -11,21 +11,22 @@ WORKDIR /build
...
 RUN find . -name 'libddwaf.so' -delete
-RUN rm -rf ./python/lib/$runtime/site-packages/urllib3
+RUN rm -rf ./python/lib/$runtime/site-packages/urllib3*
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/appsec/_iast/_taint_tracking/*.so
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/appsec/_iast/_stacktrace*.so
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/libdd_wrapper.so
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/ddup/_ddup.*.so
 RUN rm ./python/lib/$runtime/site-packages/ddtrace/internal/datadog/profiling/stack_v2/_stack_v2.*.so
 
+RUN PYTHONNODEBUGRANGES=1 python -m compileall -o 2 -b ./python/lib/$runtime/site-packages
```
duncanista, line 27: Could we add comments just to remind what's happening here?

duncanista (approved):
LGTM – would like to see this in self monitoring for a while before we release, it might be amazing to see this changes go there!

scripts/check_layer_size.sh
```diff
@@ -7,10 +7,9 @@
 
 # Compares layer size to threshold, and fails if below that threshold
 
-# 7 mb size limit
 set -e
-MAX_LAYER_COMPRESSED_SIZE_KB=$(expr 4 \* 1024)
-MAX_LAYER_UNCOMPRESSED_SIZE_KB=$(expr 13 \* 1024)
+MAX_LAYER_COMPRESSED_SIZE_KB=$(expr 5 \* 1024)
+MAX_LAYER_UNCOMPRESSED_SIZE_KB=$(expr 12 \* 1024)
```
purple4reina, line 12: Bytecode is harder to compress (hence increase from 4->5) but uncompressed is much smaller (hence decrease from 13->12)

duncanista: Sounds good! Could we update this changes in `serverless-tools` once this is merged? Thanks!

PR #522, opened 2024-10-16
feat: [SVLS-5677] DynamoDB Stream event span pointers

apiarian-datadog:
### What does this PR do?

Adds Span Pointers for Lambdas triggered by DynamoDB stream events.

### Testing Guidelines

Unit tests. Also checked with a live test lambda.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/span_pointers.py
+++ b/datadog_lambda/span_pointers.py
@@ -30,7 +30,7 @@ def calculate_span_pointers(
     except Exception as e:
         logger.warning(
             "failed to calculate span pointers for event: %s",
-            str(e),
+            e,
         )
 
     return []
@@ -75,7 +75,7 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
     except KeyError as e:
         logger.warning(
             "missing s3 information required to make a span pointer: %s",
-            str(e),
+            e,
         )
         return []
 
@@ -92,7 +92,7 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
     except Exception as e:
         logger.warning(
             "failed to generate S3 span pointer: %s",
-            str(e),
+            e,
         )
         return []
 
@@ -119,7 +119,7 @@ def _calculate_dynamodb_span_pointers_for_event_record(
     except Exception as e:
         logger.warning(
             "missing DynamoDB information required to make a span pointer: %s",
-            str(e),
+            e,
         )
         return []
 
@@ -135,7 +135,7 @@ def _calculate_dynamodb_span_pointers_for_event_record(
     except Exception as e:
         logger.warning(
             "failed to generate DynamoDB span pointer: %s",
-            str(e),
+            e,
         )
         return []
 
--- a/datadog_lambda/span_pointers.py
+++ b/datadog_lambda/span_pointers.py
@@ -152,7 +152,7 @@ def _extract_table_name_from_dynamodb_stream_record(record) -> str:
         raise ValueError(f"unexpected eventSourceARN format: {event_source_arn}")
 
     [_table, table_name, _stream, _timestamp] = dynamodb_info.split("/")
-    if not _table == "table" or not _stream == "stream":
+    if _table != "table" or _stream != "stream":
         raise ValueError(f"unexpected eventSourceARN format: {event_source_arn}")
 
     return table_name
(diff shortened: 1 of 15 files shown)


apiarian-datadog:
Will update the PR once `dd-trace-py` `2.15.0` is actually released (probably next week)

apiarian-datadog:
Confirmed that this works as expected and creates usable links with a live test lambda

datadog_lambda/span_pointers.py
```diff
@@ -89,3 +95,64 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
...
+def _extract_table_name_from_dynamodb_stream_record(record) -> str:
+    # Example eventSourceARN:
+    # arn:aws:dynamodb:us-east-2:123456789012:table/my-table/stream/2024-06-10T19:26:16.525
+    event_source_arn = record["eventSourceARN"]
+
+    [_arn, _aws, _dynamodb, _region, _account, dynamodb_info] = event_source_arn.split(
+        ":", maxsplit=5
+    )
+    if _arn != "arn" or _aws != "aws" or _dynamodb != "dynamodb":
+        raise ValueError(f"unexpected eventSourceARN format: {event_source_arn}")
```
nhulston, line 152: Do we need to check all 3? Could we just check `if _arn != "arn":`? Just for performance sake

datadog_lambda/span_pointers.py
```diff
@@ -89,3 +95,64 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
...
+
+    [_table, table_name, _stream, _timestamp] = dynamodb_info.split("/")
+    if not _table == "table" or not _stream == "stream":
```
nhulston, line 155: Do we need to check both?

Nit: `if table != "table" or _stream != "stream":` is more readable

datadog_lambda/span_pointers.py
```diff
@@ -2,7 +2,10 @@
 import logging
 from typing import List
 
-from ddtrace._trace.utils_botocore.span_pointers import (
+from ddtrace._trace.utils_botocore.span_pointers.dynamodb import (
+    _aws_dynamodb_item_span_pointer_description,
+)
+from ddtrace._trace.utils_botocore.span_pointers.s3 import (
```
duncanista, line 8: Can we avoid importing at the top? And instead, do it conditionally where we need it?

datadog_lambda/span_pointers.py
```diff
@@ -89,3 +95,64 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
...
+    record,
+) -> List[_SpanPointerDescription]:
+    try:
+        table_name = _extract_table_name_from_dynamodb_stream_record(record)
+        primary_key = record["dynamodb"]["Keys"]
+
+    except Exception as e:
+        logger.warning(
+            "missing DynamoDB information required to make a span pointer: %s",
+            str(e),
```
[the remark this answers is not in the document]
apiarian-datadog: o, ok, i'll remove the str wrappers!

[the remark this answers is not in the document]
apiarian-datadog: removed.

datadog_lambda/span_pointers.py
```diff
@@ -89,3 +95,64 @@ def _calculate_s3_span_pointers_for_object_created_s3_information(
             str(e),
```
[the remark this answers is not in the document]
apiarian-datadog: removed.

datadog_lambda/span_pointers.py
apiarian-datadog: we don't have to check all three, but i prefer to be super cautious here.

datadog_lambda/span_pointers.py
apiarian-datadog: ah, i'll flip it around!

apiarian-datadog: done

datadog_lambda/span_pointers.py
apiarian-datadog: you mean move this function import down to where we're about to call the function in `_calculate_s3_span_pointers_for_object_created_s3_information`? we could, but that's a pretty funky thing to do in python. what would be the benefit? it may harm readability.

PR #524, opened 2024-10-25
Add `http.route` tags for API Gateway

nhulston:
### What does this PR do?
Adds `http.route` tags for Lambdas that were triggered by API Gateway.
- For APIGW v1, we set the route to the resource
- For APIGW v2, we set the route to the routeKey

### Motivation

These tags are used by API Catalog
https://datadoghq.atlassian.net/browse/SVLS-5779

A similar PR for universal instrumentation (Java, .NET, Golang) is already done: https://github.com/DataDog/datadog-agent/commit/315c38f20f75a4a5ce90fcf691d10d3ed286966d

See https://github.com/DataDog/datadog-lambda-js/pull/584 for JavaScript changes.

### Testing Guidelines

Manual tests; update unit tests and snapshots

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -302,6 +302,7 @@ def extract_http_tags(event):
     # Try to get `routeKey` from API GW v2; otherwise try to get `resource` from API GW v1
     route = event.get("routeKey") or event.get("resource")
     if route:
+        # "GET /my/endpoint" = > "/my/endpoint"
         http_tags["http.route"] = route.split(" ")[-1]
 
     return http_tags
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -299,6 +299,11 @@ def extract_http_tags(event):
     if headers and headers.get("Referer"):
         http_tags["http.referer"] = headers.get("Referer")
 
+    # Try to get `routeKey` from API GW v2; otherwise try to get `resource` from API GW v1
+    route = event.get("routeKey") or event.get("resource")
+    if route:
+        http_tags["http.route"] = route
+
     return http_tags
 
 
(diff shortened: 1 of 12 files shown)


purple4reina (approved):

datadog_lambda/trigger.py
```diff
@@ -299,6 +299,11 @@ def extract_http_tags(event):
     if headers and headers.get("Referer"):
         http_tags["http.referer"] = headers.get("Referer")
 
+    # Try to get `routeKey` from API GW v2; otherwise try to get `resource` from API GW v1
+    route = event.get("routeKey") or event.get("resource")
+    if route:
```
duncanista, line 304: Do we really need to define a variable here?

duncanista: I'm pretty sure we could just have a `None` tag and it should be ignored by the backend, but I'm not sure how risky is that, I'd check the tracer

duncanista (approved):

tests/integration/snapshots/logs/async-metrics_python310.log
```diff
@@ -605,6 +606,7 @@ HTTP GET https://www.datadoghq.com/ Headers: ["Accept-Encoding:gzip, deflate","A
           "http.url": "XXXX.execute-api.eu-west-1.amazonaws.com",
           "http.url_details.path": "/httpapi/get",
           "http.method": "GET",
+          "http.route": "GET /httpapi/get",
```
wantsui, line 609: Should this http.route span tag include `GET`? Looking at the Datadog Agent implementation, it leaves out the method: https://github.com/DataDog/datadog-agent/commit/315c38f20f75a4a5ce90fcf691d10d3ed286966d .

nhulston: Good catch!

datadog_lambda/trigger.py
nhulston: We have to parse the route now, so I think it makes sense to leave the variable now.
`http_tags["http.route"] = route.split(" ")[-1]`

wantsui (approved):
I defer the serverless portion of the review to @duncanpharvey or anyone else in the Serverless side of things but the test results of `http.route` look good to me.

PR #526, opened 2024-10-29
Explicit trace ID propagation for SFN

avedmala:
### What does this PR do?

Add logic to extract trace context from `_datadog` for Step Functions cases where...
1. Root of the overall execution is a Lambda. This means we have the `x-datadog-trace-id` but we need to compute the `x-datadog-parent-id` using `x-datadog-parent-id-hash`
2. Root of the overall execution is another Step Function. This means we need to compute both `x-datadog-trace-id` and `x-datadog-parent-id` from their respective hashes
3. Keeping legacy code where there's no `_datadog` and we instead figure out context from the Execution and State info

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tracing.py
```diff
@@ -380,10 +381,26 @@ def extract_context_from_step_functions(event, lambda_context):
...
-        # https://github.com/DataDog/dd-trace-py/blob/3e34d21cb9b5e1916e549047158cb119317b96ab/ddtrace/propagation/http.py#L232-L240
-        trace_id = _deterministic_sha256_hash(execution_id, LOWER_64_BITS)
+        meta = {}
+
+        if "_datadog" in event:
+            # use the trace ID from the top-most parent when it exists
+            trace_header = event.get("_datadog")
+            trace_id = trace_header.get(TraceHeader.TRACE_ID)
+            tags = trace_header.get(TraceHeader.TAGS)
+            for tag in tags.split(","):
```
avedmala, line 391: This way we pass all the tags and not just `_dd.p.tid`, not sure if there's any unintended consequences

kimi-p (commented):
Can you test with Lambda1 -> Lambda2 with payload of `_datadog` and see if these two spans will link? You might need to build the python layer and use it. Thanks!

This is the command that joey shared to me when I was building py layer
```

# python
docker buildx build -t datadog-lambda-python-amd64:3.9 . --no-cache \
        --build-arg image=python:3.9 \
        --build-arg runtime=python3.9 \
        --platform linux/amd64 \
        --progress=plain \
        -o ./.layers/tmp/python && \
pushd ./.layers/tmp && zip -q -r datadog_lambda_py-amd64-3.9.zip ./ && \
sso_sand_run aws lambda publish-layer-version \
 --layer-name "Datadog-Python39" \
 --region sa-east-1 \
 --zip-file  "fileb://datadog_lambda_py-amd64-3.9.zip" && popd
 ```

**Review by kimi-p — APPROVED** [review]

File: datadog_lambda/tracing.py [hunk 2]
```diff
@@ -380,10 +381,33 @@ def extract_context_from_step_functions(event, lambda_context):
...
-        # traditional traceId and _dd.p.tid tag
-        # https://github.com/DataDog/dd-trace-py/blob/3e34d21cb9b5e1916e549047158cb119317b96ab/ddtrace/propagation/http.py#L232-L240
-        trace_id = _deterministic_sha256_hash(execution_id, LOWER_64_BITS)
+        meta = {}
+
+        if "_datadog" in event:
+            trace_header = event.get("_datadog")
+            if TraceHeader.TRACE_ID in trace_header:
+                # use the trace ID from the top-most parent when it exists
+                trace_id = int(trace_header.get(TraceHeader.TRACE_ID))
```
joeyzhao2018, line 390: Why don't we use the `propagator.extract` pattern here?

joeyzhao2018 (changes requested):
I'm not convinced that this is any exception for `propagator.extract`.
I'll update my proposed changes later.

datadog_lambda/tracing.py
```diff
@@ -380,10 +381,33 @@ def extract_context_from_step_functions(event, lambda_context):
...
+                for tag in tags.split(","):
+                    tag_key, tag_val = tag.split("=")
+                    meta[tag_key] = tag_val
+            elif "x-datadog-execution-arn" in trace_header:
+                root_execution_id = trace_header.get("x-datadog-execution-arn")
+                trace_id = _deterministic_sha256_hash(root_execution_id, LOWER_64_BITS)
+                meta["_dd.p.tid"] = hex(
+                    _deterministic_sha256_hash(root_execution_id, HIGHER_64_BITS)
+                )[2:]
+        else:
```
joeyzhao2018, line 401:
```suggestion
       trace_id = None
        if "_datadog" in event:
            trace_header = event.get("_datadog")
            explicit_context = propagator.extract(trace_header)
            if _is_context_complete(explicit_context):
                tags = trace_header.get(TraceHeader.TAGS, "")
                for tag in tags.split(","):
                    tag_key, tag_val = tag.split("=")
                    meta[tag_key] = tag_val
                # I have a question here, is "_dd.p.tid" already set here?

                explicit_context.span_id = _deterministic_sha256_hash(f"{execution_id}#{state_name}#{state_entered_time}", HIGHER_64_BITS)
                return explicit_context
            
            if "x-datadog-execution-arn" in trace_header:
                root_execution_id = trace_header.get("x-datadog-execution-arn")
                trace_id = _deterministic_sha256_hash(root_execution_id, LOWER_64_BITS)
                meta["_dd.p.tid"] = hex(
                    _deterministic_sha256_hash(root_execution_id, HIGHER_64_BITS)
                )[2:]

        if trace_id is None:
```

datadog_lambda/tracing.py
```diff
@@ -357,8 +358,10 @@ def extract_context_from_kinesis_event(event, lambda_context):
 
 
 def _deterministic_sha256_hash(s: str, part: str) -> (int, int):
-    sha256_hash = hashlib.sha256(s.encode()).hexdigest()
+    return _sha256_to_binary_part(hashlib.sha256(s.encode()).hexdigest(), part)
```
avedmala, line 361: Split this up with a helper so we have an entrypoint for our `sha256_hash` from the `x-datadog-trace-id-hash` and `x-datadog-parent-id-hash`

datadog_lambda/tracing.py
```diff
@@ -377,30 +380,53 @@ def extract_context_from_step_functions(event, lambda_context):
...
+            else:  # sfn root
+                trace_id = _sha256_to_binary_part(
+                    dd_data.get("x-datadog-trace-id-hash"), LOWER_64_BITS
+                )
+                meta["_dd.p.tid"] = hex(
+                    _sha256_to_binary_part(
+                        dd_data.get("x-datadog-trace-id-hash"), HIGHER_64_BITS
+                    )
+                )[2:]
+        else:
```
avedmala, line 406: This whole branch should be the old behavior

datadog_lambda/tracing.py
joeyzhao2018: nice!

datadog_lambda/tracing.py
```diff
@@ -377,30 +379,49 @@ def extract_context_from_step_functions(event, lambda_context):
...
-
-        parent_id = _deterministic_sha256_hash(
-            f"{execution_id}#{state_name}#{state_entered_time}", HIGHER_64_BITS
-        )
+        meta = {}
+
+        if "_datadog" in event:
+            dd_data = event.get("_datadog")
+            parent_id = _sha256_to_binary_part(
+                dd_data.get("x-datadog-parent-id-hash"), HIGHER_64_BITS
```
kimi-p, line 387: Note: Since this `extract_context_from_step_functions()` is executed when we know `_datadog.serverless-version=v2`, we don't need to check if this hash key exists. The v2 contract is that they need to have the `_datadog.x-datadog-parent-id-hash`.

kimi-p (approved):
LGTM. Let's test this change with reducer change in staging before merge it.

datadog_lambda/tracing.py
avedmala: I might be misunderstanding, we're not checking if the hash key exists right?

We're computing the `parent_id` no matter what, the only thing we check in this branch is if `x-datadog-trace-id` exists

avedmala: So we're saying it's the same `EventTypes.STEP_FUNCTIONS` whether its the old format with context object or new one with `_datadog`. Both cases will enter `extract_context_from_step_functions()`

If we want to separate them, one thing we could do is make a new V2 event with a V2 extractor to go along with it

Issue #532, opened 2024-11-05
'_LambdaDecorator' object has no attribute 'trigger_tags'


CatalinAndrei-Flam:
## Expected Behavior
Datadog should not throw errors, in previous version this did not happen. (extension_layer_version:64, python_layer_version:98)

## Actual Behavior
Datadog throws errors because it cannot extract the tags from the event. In this case the incoming event is not a dictionary but a list of dictionaries that's generated by a dynamodb stream in an aws eventbridge pipe with batching enabling. 

## Steps to Reproduce the Problem

  1. Update extension_layer_version to 65 and python_layer_version to 99
  2. Create. eventbridge pipe with dynamodb stream to lambda  and batching enabled .
  3. Trigger multiple changes in the dynamodb table with stream enabled which will generate a list of dicts as an input event for the lambda that's failing.
![Image](https://github.com/user-attachments/assets/eaad450d-4d98-40be-ad0b-4960d7e7d5d4)

## Specifications

  - extension_layer_version": 65
  - python_layer_version: 99
  - Python version: 3.11
  
## Stacktrace
  ```
  [ERROR]	2024-10-30T07:48:05.533Z	faa1c31f-9cde-4738-ba9d-fd118ea51f8e	Error 'list' object has no attribute 'get'. Traceback: Traceback (most recent call last):
  File "./python/lib/python3.11/site-packages/datadog_lambda/wrapper.py", line 282, in _before
  File "./python/lib/python3.11/site-packages/datadog_lambda/tracing.py", line 414, in is_legacy_lambda_step_function
AttributeError: 'list' object has no attribute 'get'
  ```
and

  ```
[ERROR]	2024-10-30T07:48:05.323Z	aca884cf-00d7-4b80-8764-6883788465f2	Error '_LambdaDecorator' object has no attribute 'trigger_tags'. Traceback: Traceback (most recent call last):
  File "./python/lib/python3.11/site-packages/datadog_lambda/wrapper.py", line 331, in _after
AttributeError: '_LambdaDecorator' object has no attribute 'trigger_tags'
  ```

purple4reina:
Thank you for reporting this. We are looking to get a fix out right away.

purple4reina:
Hi @CatalinAndrei-Flam, we have a released layer v101 which contains a fix for this issue. Please test it out and let us know if there's anything else you need.

anthonyangel:
I'm seeing a regression and this is being caused again in layer version 105

PR #533, opened 2024-11-05
fix: add typecheck to lambda wrapper

avedmala:
### What does this PR do?

* Checks if `event` is a dictionary before trying to call `get()` on it as it may be a list
* The case where it is a list is breaking instrumentation for customers
* Extended unit tests to catch this

Closes https://github.com/DataDog/datadog-lambda-python/issues/532

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -411,6 +411,9 @@ def is_legacy_lambda_step_function(event):
     """
     Check if the event is a step function that called a legacy lambda
     """
+    if not isinstance(event, dict):
+        return False
+
     event = event.get("Payload", {})
     return "Execution" in event and "StateMachine" in event and "State" in event
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -411,10 +411,10 @@ def is_legacy_lambda_step_function(event):
     """
     Check if the event is a step function that called a legacy lambda
     """
-    if not isinstance(event, dict):
+    if not isinstance(event, dict) or "Payload" not in event:
         return False
 
-    event = event.get("Payload", {})
+    event = event.get("Payload")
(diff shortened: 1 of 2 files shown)


purple4reina (approved):

datadog_lambda/tracing.py
```diff
@@ -411,6 +411,9 @@ def is_legacy_lambda_step_function(event):
     """
     Check if the event is a step function that called a legacy lambda
     """
+    if not isinstance(event, dict):
```
apiarian-datadog, line 414: do we have other code that assumes that this is a dict in the wrapper flow that will trip over this problem?

also, should we look at the contents of the list if it is a list? or is that impossible for a legacy lambda step function?

purple4reina: I dont' think we need to look through the whole list. If it's not a dict, we know right away that it's not a step function event.

As for the isinstance, we do have other checks for if the event is a dict. For example in determining trace propagation we skip altogether if it's not a dict. Best answer could even be surrounding the whole thing with a try/except.

datadog_lambda/tracing.py
```diff
@@ -411,6 +411,9 @@ def is_legacy_lambda_step_function(event):
...
+        return False
+
     event = event.get("Payload", {})
```
purple4reina, line 417: This isn't super related to this incident, but I think it's worth changing anyway. I'm wondering if we can exit early if `"Payload"` is not in the event. By then we already know that it's not gonna be a step function event. So, exiting early would prevent the allocation of the `{}` and all the searching on the next line.

avedmala: Good callout 👍🏽 , will add an early exit

purple4reina (approved):

PR #537, opened 2024-11-12
Explicit trace ID propagation for SFN w/o Hashing

avedmala:
### What does this PR do?
Forked off of https://github.com/DataDog/datadog-lambda-python/pull/526, this PR will use context objects to generate traceIDs rather than generating them form hashes

Add logic to extract trace context from `_datadog` for Step Functions cases where...
1. Root of the overall execution is a Lambda. This means we have the `x-datadog-trace-id` but we need to compute the `x-datadog-parent-id` using parent SFN context object
2. Root of the overall execution is another Step Function. This means we need to compute both `x-datadog-trace-id` and `x-datadog-parent-id` from `RootExecutionId` and the parent SFN context object
3. Keeping legacy code where there's no `_datadog` and we instead figure out context from the Execution and State info

Also fixed the trigger tags parsing for our use case!

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tracing.py
```diff
@@ -415,7 +465,10 @@ def is_legacy_lambda_step_function(event):
         return False
 
     event = event.get("Payload")
-    return "Execution" in event and "StateMachine" in event and "State" in event
+    return isinstance(event, dict) and (
```
avedmala, line 468: Updates the legacy lambda check to account for `_datadog` case and added the extra check for safety

datadog_lambda/tracing.py
```diff
@@ -371,36 +370,87 @@ def _deterministic_sha256_hash(s: str, part: str) -> (int, int):
...
+    """
+    if trace_tags:
+        for tag in trace_tags.split(","):
+            if "_dd.p.tid=" in tag:
+                return tag.split("=")[1]
+
+    return ""
+
+
+def _generate_sfn_parent_id(context: dict) -> int:
```
avedmala, line 386: Created these helpers to pull hashing and hex related details out of `extract_context_from_step_functions` as it was getting pretty verbose

datadog_lambda/tracing.py
```diff
@@ -371,36 +370,88 @@ def _deterministic_sha256_hash(s: str, part: str) -> (int, int):
...
-        # https://github.com/DataDog/dd-trace-py/blob/3e34d21cb9b5e1916e549047158cb119317b96ab/ddtrace/propagation/http.py#L232-L240
-        trace_id = _deterministic_sha256_hash(execution_id, LOWER_64_BITS)
-
-        parent_id = _deterministic_sha256_hash(
-            f"{execution_id}#{state_name}#{state_entered_time}", HIGHER_64_BITS
-        )
+        meta = {}
+        dd_data = event.get("_datadog")
+
+        if dd_data and dd_data.get("serverless-version") == "v2":
```
nine5two7, line 427: What is the difference between v1 and v2? Should we start from version v2 or v1?

avedmala: There's no v1, I think we can start from v1 instead if we want

I think @kimi-p suggested v2, maybe because the "legacy" context is implied to be v1

nine5two7: In my opinion, "V2" may be somewhat confusing to customers because there is no explicit reference to "V1" at all.

kimi-p: sgtm!

datadog_lambda/tracing.py
kimi-p: If a customer use Legacy Lambda and legacy schema (meaning not using v1), they won't have `_datadog`.

datadog_lambda/trigger.py
```diff
@@ -254,6 +256,13 @@ def parse_event_source_arn(source: _EventSource, event: dict, context: Any) -> s
     if source.event_type == EventTypes.CLOUDWATCH_EVENTS and event.get("resources"):
         return event.get("resources")[0]
 
+    # e.g. arn:aws:states:us-east-1:123456789012:stateMachine:stateMachineName
```
kimi-p, line 259:
```suggestion
    # Returning state machine arn as event source arn.
```

kimi-p (commented):
One question regarding the legacy lambda check that I think we need to update, otherwise LGTM! Thank you!

kimi-p (approved):
@joeyzhao2018 Could you take a look to confirm that we are using the right extraction pattern in this PR?

PR #545, opened 2024-12-04
Redriven Step Functions Trace Merging

avedmala:
### What does this PR do?

Adds support for Step Functions trace merging in Redrive cases

We previously used `hash(ExecutionId # StateName # StateEnteredTime)` for spanID calculation but these values are identical across redrives for a Lambda task state. The new approach also adds a `RedriveCount` to the end of the hash but omits this value when it is 0 to have easy backwards compatability.

### Motivation

https://github.com/DataDog/logs-backend/pull/84188

### Testing Guidelines

Redriven [Trace](https://dd.datad0g.com/apm/trace/5be190b56606251d6deb0f8496924f74?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=9522989320986969024&timeHint=1734368341071) with Lambda and Trace Merging
Screenshot 2024-12-16 at 12 00 25 PM

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -392,7 +392,7 @@ def _generate_sfn_parent_id(context: dict) -> int:
     version of the Lambda layer that doesn't use this value for its parentID generation.
     """
     execution_id = context.get("Execution").get("Id")
-    redrive_count = context.get("Execution").get("RedriveCount")
+    redrive_count = context.get("Execution").get("RedriveCount", 0)
     state_name = context.get("State").get("Name")
     state_entered_time = context.get("State").get("EnteredTime")
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -385,11 +385,11 @@ def _parse_high_64_bits(trace_tags: str) -> str:
 
 def _generate_sfn_parent_id(context: dict) -> int:
     """
-    The upstream Step Function can propagate its execution context to downstream Lambdas. The Lambda can use these
-    details to share the same traceID and infer its parent's spanID.
+    The upstream Step Function can propagate its execution context to downstream Lambdas. The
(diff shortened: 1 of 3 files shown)


datadog_lambda/tracing.py
```diff
@@ -384,12 +384,23 @@ def _parse_high_64_bits(trace_tags: str) -> str:
...
+
+    Excluding redriveCount when its 0 to account for cases where customers are using an old
+    version of the Lambda layer that doesn't use this value for its parentID generation.
+    """
     execution_id = context.get("Execution").get("Id")
+    redrive_count = context.get("Execution").get("RedriveCount")
     state_name = context.get("State").get("Name")
     state_entered_time = context.get("State").get("EnteredTime")
 
+    redrive_postfix = "" if redrive_count == 0 else f"#{redrive_count}"
```
nine5two7, line 399: Is it possible that redrive _count is null?

avedmala: `Execution.RedriveCount` will **always** be in the SFNs context object, defaulting to 0

Only `Execution.RedriveTime` has a risk of being missing, only appearing when redrive count is 1 or more

https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html

avedmala: Even though this is supposed to be true, I found some instances in customer logs where RedriveCount is null. Wasn't able to reproduce it myself. I'll add some null-check here and also for the node layer change accordingly

kimi-p: I guess we'll need to assume all the new fields could be `null` in the future. There may be some edge cases that AWS does not handle well.

kimi-p (approved):
lgtm

PR #546, opened 2024-12-19
feat: use vendored version of requests package in layer.

purple4reina:
### What does this PR do?

Removes the requests package from the zipped lambda layer. Adds `/path/to/pip/_vendor` to sys.path so requests can instead be imported from `pip._vendor.requests`.

### Motivation

Removing requests and all its dependencies from the layer reduces the package size (unzipped) from 15,212 KB to 13,904 KB -- a roughly **8.5% reduction**!! (Data gathered running `PYTHON_VERSION=3.12 ARCH=arm64 ./scripts/build_layers.sh`)

### Potential Drawbacks

Some customers may already be relying on the presence of the `requests` package in our python layer. This change could crash their lambda functions if `requests` is imported before `datadog_lambda`. This would only be the case if they are using the `@datadog_lambda_wrapper` decorator and not the automatic handler redirection using `DD_LAMBDA_HANDLER`. 

For example, prior to this change, assuming a customer adds just our python layer and nothing else, this handler will fail to initialize.

```py
import requests

from datadog_lambda.wrapper import datadog_lambda_wrapper

@datadog_lambda_wrapper
def handler(event, context):
    resp = requests.get('https://example.com')
    resp.raise_for_status()
    return 'ok'
```

Therefore, we absolutely must include something about the import order importance in the release notes.

### Additional Notes

If for some reason the importing of `requests` doesn't work (say `pop._vendor.requests` isn't installed on the container) then the only situation in which the lambda function would raise an exception is when attempting to submit metrics via the datadog api. We do this when either the metric has a post dated timestamp or there is no extension available. Therefore, a pretty narrow use case.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


tests/integration/handle.py
```diff
@@ -21,6 +22,8 @@ def handle(event, context):
 
     # Generate custom metrics
     lambda_metric("hello.dog", 1, tags=["team:serverless", "role:hello"])
+    lambda_metric("hello.cat", 1, tags=["team:serverless", "role:hello"],
+        timestamp=time.time() - 60)
```
purple4reina, line 26: Creating a custom metric with a timestamp forces our instrumentation to use the `datadog` api package instead of sending the metric to the extension.

tests/integration/handle.py
```diff
@@ -20,7 +21,11 @@ def handle(event, context):
             record_ids.append(record["Sns"]["MessageId"])
 
     # Generate custom metrics
+    timestamp = time.time() - 60
     lambda_metric("hello.dog", 1, tags=["team:serverless", "role:hello"])
+    lambda_metric(
+        "hello.cat", 1, tags=["team:serverless", "role:hello"], timestamp=timestamp
+    )
```
purple4reina, line 28: Creating a custom metric with a timestamp forces our instrumentation to use the datadog api package instead of sending the metric to the extension.

purple4reina:
We have tried this idea before, though using the requests version vendored with botocore. However there were major problems and we reverted. See

* https://github.com/DataDog/datadog-lambda-python/pull/22
* https://github.com/DataDog/datadog-serverless-functions/pull/424

Since we do not have any control over the lambda runtime, if for some reason the requests version changes or goes away, we have little recourse in trying to solve it. There is potential that it could cause every lambda everywhere to start failing, which is an incredibly high risk, even if chances of it happening are very low.

PR #565, opened 2025-02-27
Release with ddtrace@2.21.3

joeyzhao2018:
### What does this PR do?

Release with ddtrace@2.21.3

### Motivation

Include a fix for payload tagging for dynamodb

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN PYTHONNODEBUGRANGES=1 python -OO -m compileall -b ./python/lib/$runtime/site
 # remove all .py files except ddtrace/contrib/*/__init__.py which are necessary
 # for ddtrace.patch to discover instrumationation packages.
 RUN find ./python/lib/$runtime/site-packages -name \*.py | grep -v ddtrace/contrib | xargs rm -rf
-RUN find ./python/lib/$runtime/site-packages/ddtrace/contrib -name \*.py | grep -v __init__ | xargs rm -rf
+# RUN find ./python/lib/$runtime/site-packages/ddtrace/contrib -name \*.py | grep -v __init__ | xargs rm -rf
```
purple4reina, line 53: Commenting out this line will have the impact of increasing the layer size significantly. We shouldn't do this.

purple4reina: It will also have the impact of increasing cold start time.

joeyzhao2018: I'm aware of the related drawbacks. First comment it out to see it's related as the integration tests passed. 
The specific error complained about `logging` module. So maybe we just keep the `logging`'s py files?

joeyzhao2018: Updated.

Dockerfile
```diff
@@ -50,7 +50,9 @@ RUN PYTHONNODEBUGRANGES=1 python -OO -m compileall -b ./python/lib/$runtime/site
 # remove all .py files except ddtrace/contrib/*/__init__.py which are necessary
 # for ddtrace.patch to discover instrumationation packages.
 RUN find ./python/lib/$runtime/site-packages -name \*.py | grep -v ddtrace/contrib | xargs rm -rf
-RUN find ./python/lib/$runtime/site-packages/ddtrace/contrib -name \*.py | grep -v __init__ | xargs rm -rf
+RUN find ./python/lib/$runtime/site-packages/ddtrace/contrib -name "*.py" \
+    -not -path "./python/lib/$runtime/site-packages/ddtrace/contrib/internal/logging/*" \
+    | grep -v __init__ | xargs rm -rf
```
purple4reina, line 55: Unfortunately, this isn't the right fix. To better understand the problem, take a look at this line change in this PR: https://github.com/DataDog/dd-trace-py/pull/12153/files#diff-6a70144455f9fea41df44b3e578cfaa81ae9c5297e41cce9ef02a5b719013817R265

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 python = ">=3.8.0,<4"
 datadog = ">=0.51.0,<1.0.0"
 wrapt = "^1.11.2"
-ddtrace = ">=2.20.0"
+ddtrace = "==2.21.3"
```
purple4reina, line 31: Hmm, one concern here is that this change will also impact customers not using the layer. 

I'm wondering if instead of changing the compatibility requirements here, we should just force an install of the specific ddtrace version in the Dockerfile.

purple4reina: So basically, here https://github.com/DataDog/datadog-lambda-python/blob/b64be3b098faf2743af1a1009f0c414f7fae8f33/Dockerfile#L17, we'd change this to something like

```Dockerfile
RUN pip install . ddtrace==2.21.3 -t ./python/lib/$runtime/site-packages
```

purple4reina: Actually, I'm fine with however you wanna handle this @joeyzhao2018 since we're just gonna revert it back right away afterwards.

joeyzhao2018: The docker file is a better place i agree.

purple4reina (approved):

PR #572, opened 2025-03-10
Enable LLM Observability with `agentless_enabled=True` by default with a parsed API key

sabrenner:
### What does this PR do?

Enables LLM Observability with `agentless_enabled=True` to ensure seamless compatibility with the agent used in the DD Extension layer. Previously, LLM Observability would try and use an agent proxy endpoint which doesn't exist on the trace agent used in the `next` version of the DD Extension layer. 

Since all we really use the proxy for outside of serverless is so that users don't have to re-state their API key, it should be fine to just use agentless for serverless environments by default (as LLM Observability now also still sends APM traces even if `agentless_enabled=True` is set).

To help enable this, I added a call to `init_api` to get the `DD_API_KEY` from the secrets manager if it lives there, to make the experience even smoother. Otherwise, we can enforce `DD_API_KEY` in the lambda function's env vars

### Motivation

MLOB-2225

### Testing Guidelines

Built the layer with the `build_layers` script, and verified against our LLM Observability lambda functions with only
- `DD_LLMOBS_ENABLED`
- `DD_LLMOBS_ML_APP`
- `DD_API_KEY`

set, and verified traces showed up in the UI.

### Additional Notes

Happy to remove the `init_api` part if it would be too much of a burden on the code path for a serverless env. I saw it was available, and only used for metrics, but decided to re-use. It can be revisited later if needed and we can enforce `DD_API_KEY` being set directly instead.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR impacts documentation, and it has been updated (or a ticket has been logged)

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -56,9 +56,15 @@ profiling_env_var = os.environ.get("DD_PROFILING_ENABLED", "false").lower() == "
 if profiling_env_var:
     from ddtrace.profiling import profiler
 
+llmobs_api_key = None
 llmobs_env_var = os.environ.get("DD_LLMOBS_ENABLED", "false").lower() in ("true", "1")
 if llmobs_env_var:
+    from datadog_lambda.api import init_api
+    from datadog import api
     from ddtrace.llmobs import LLMObs
+
+    init_api()
+    llmobs_api_key = api._api_key
 
 logger = logging.getLogger(__name__)
 
@@ -229,7 +235,10 @@ class _LambdaDecorator(object):
 
             # Enable LLM Observability
             if llmobs_env_var:
-                LLMObs.enable()
+                LLMObs.enable(
+                    agentless_enabled=True,
+                    api_key=llmobs_api_key,
+                )
 
             logger.debug("datadog_lambda_wrapper initialized")
         except Exception as e:
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -59,12 +59,10 @@ if profiling_env_var:
 llmobs_api_key = None
 llmobs_env_var = os.environ.get("DD_LLMOBS_ENABLED", "false").lower() in ("true", "1")
 if llmobs_env_var:
-    from datadog_lambda.api import init_api
-    from datadog import api
+    from datadog_lambda.api import get_api_key
     from ddtrace.llmobs import LLMObs
 
-    init_api()
-    llmobs_api_key = api._api_key
+    llmobs_api_key = get_api_key()
 
 logger = logging.getLogger(__name__)
 
--- a/datadog_lambda/api.py
+++ b/datadog_lambda/api.py
@@ -50,7 +50,8 @@ def decrypt_kms_api_key(kms_client, ciphertext):
 def get_api_key() -> str:
     """
     Gets the Datadog API key from the environment variables or secrets manager.
-    Extracts the result to a global value to avoid repeated calls to the secrets manager from different products.
+    Extracts the result to a global value to avoid repeated calls to the
+    secrets manager from different products.
     """
     global api_key
     if api_key:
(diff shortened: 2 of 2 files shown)


datadog_lambda/wrapper.py
```diff
@@ -56,10 +56,16 @@
 if profiling_env_var:
     from ddtrace.profiling import profiler
 
+llmobs_api_key = None
 llmobs_env_var = os.environ.get("DD_LLMOBS_ENABLED", "false").lower() in ("true", "1")
 if llmobs_env_var:
+    from datadog_lambda.api import init_api
```
astuyve, line 62: nice use of lazy loading here!

astuyve (commented):
This looks pretty good to me, a few things to keep in mind:
- if LLMobs ships data agentlessly via nonblocking thread or async runtime, you should expect to drop or miss data as the CPU is throttled and resumed between invocations. You may want to check this by performing a few invokes, then waiting several minutes. You can kind of handle this by aggressively retrying, but it's not perfect.
- Because you're doing this in the main function process and not the extension, data will be lost if it's not flushed before the sandbox shuts down – as the main process is not executed on the `shutdown` event (but extensions are).
- You should also expect to lose data in the event of a timeout, OOM, or other unhandled exceptions which cause the python process to re-initialize.

I'm not a codeowner of this project any longer, so I'll let that team chime in.

datadog_lambda/wrapper.py
```diff
@@ -56,10 +56,16 @@
...
+    from datadog import api
     from ddtrace.llmobs import LLMObs
 
+    init_api()
+    llmobs_api_key = api._api_key
```
purple4reina, line 67: Do you only need the api key here, or are we actually needing to load all of `datadog`?

sabrenner: we just need the api key, but from what I can tell this `init_api` function sets the api key on `api._api_key` (and imports `api` from `datadog` itself). is there a better way I could go about getting the api key?

purple4reina: Yes, we should avoid importing `datadog` if at all possible. It's a big package and eats up a ton of time just to import it.

I suggest pulling out the code that finds and sets the api key into its own function. Then, instead of importing `init_api`, you can just import and call that single function. https://github.com/DataDog/datadog-lambda-python/blob/fc0beaae643d59a4745ae6d8da163ad39cdb5970/datadog_lambda/api.py#L57-L78

duncanista (approved):
Looks good to me!

After pairing on this, we discovered the impact in overhead for LLMObs customers.
The ballpark overhead increase for this use case would be 200ms (`datadog.api` import and secrets manager call).

So approximately ~800ms, as opposed to ~550ms.

LLMObs adds ~60ms during init.

datadog_lambda/wrapper.py
sabrenner: that makes sense to me - let me extract that part out!

sabrenner: attempted in https://github.com/DataDog/datadog-lambda-python/pull/572/commits/b2cc1ba781a39fa4ebfb6e29704d42d956107815 - let me know what you think! i hoisted the API key to a global so we could re-use it to save a little time. I do see the import calls to `datadog` completely gone now (which looked to be maybe 40-60ms).

sabrenner:
Thanks for the callouts @astuyve! 

> if LLMobs ships data agentlessly via nonblocking thread or async runtime, you should expect to drop or miss data as the CPU is throttled and resumed between invocations. You may want to check this by performing a few invokes, then waiting several minutes. You can kind of handle this by aggressively retrying, but it's not perfect.

Our LLMObs writer just re-uses the periodic HTTP writer service in `dd-trace-py`, which I think uses a synchronous blocking http request. It does run in a thread I think ( I haven't looked too deeply into the `ddtrace` internals), but I'll keep a note of this to see if it becomes an issue. I don't think our customers using LLMObs in serverless environments have too large of use cases to be hitting some kind of throttling. I'll run the tests, though, and follow up if I can make any improvements!

> Because you're doing this in the main function process and not the extension, data will be lost if it's not flushed before the sandbox shuts down – as the main process is not executed on the shutdown event (but extensions are).

We [force flush](https://github.com/DataDog/datadog-lambda-python/blob/fc0beaae643d59a4745ae6d8da163ad39cdb5970/datadog_lambda/wrapper.py#L389-L390) LLMObs in the `datadog_lambda.wrapper._LambdaDecorator._after` function, so I think we should be good in this case. This was something that got us in the past so we made sure to have that there 👍 

> You should also expect to lose data in the event of a timeout, OOM, or other unhandled exceptions which cause the python process to re-initialize.

This is a good point, which I think falls outside of this PR. I'll make a note to follow up on investigating this!

PR #573, opened 2025-03-11
Propagate Step Function Trace Context through Managed Services

avedmala:
### What does this PR do?

Allows us to extract Step Function trace context in the following cases
1. SFN -> EventBridge -> Lambda
2. SFN -> EventBridge -> SQS -> Lambda
3. SFN -> SQS -> Lambda
4. SFN -> SNS -> Lambda
5. SFN -> SNS -> SQS -> Lambda

### Motivation

Customer feature request for the following use case
* Step Function has EventBridge PutEvents task that emits an event with a task token
  * Enable “Wait for callback” on this
* EventBridge matching rule will trigger a Lambda that will handle the event and resume the Step Function with the task token

It's not much more work to add the other cases so figured might as well

### Testing Guidelines

1. SFN -> EventBridge -> Lambda [trace](https://dd.datad0g.com/apm/trace/39d8525a4963f9df2663aeedb905c2d8?colorByAttr=service&graphType=flamegraph&panel_tab=flamegraph&shouldShowLegend=true&sort=time&spanID=15873715362839237600&timeHint=1741726680739)
Screenshot 2025-03-12 at 9 51 53 AM

2. SFN -> EventBridge -> SQS -> Lambda [trace](https://dd.datad0g.com/apm/trace/73a4900811d7fb4f095e2a11e058cdfd?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4251483085491795259&timeHint=1741804286571.9998)
Screenshot 2025-03-13 at 11 33 24 AM

3. SFN -> SQS -> Lambda [trace](https://dd.datad0g.com/apm/trace/62691f1d143e15444ae19fd510e40425?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=10989714857593541278&timeHint=1741878429686)
Screenshot 2025-03-13 at 11 08 51 AM

4. SFN -> SNS -> Lambda [trace](https://dd.datad0g.com/apm/trace/3a4fd1a254eb514a34fca3bd2ab5d589?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4257047554942861127&timeHint=1741878109738)
Screenshot 2025-03-13 at 11 03 00 AM

5. SFN -> SNS -> SQS -> Lambda [trace](https://dd.datad0g.com/apm/trace/45457f5f3fde3fa14f3820e0bc98d7b8?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=3955897650210673755&timeHint=1741879766144)
Screenshot 2025-03-13 at 11 31 16 AM

### Additional Notes

The instrumentation involves injecting the `_datadog: {...}` trace context into these managed services from the Step Function definition, will add instructions in our public docs

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -349,8 +349,12 @@ def extract_context_from_eventbridge_event(event, lambda_context):
         if not dd_context:
             return extract_context_from_lambda_context(lambda_context)
 
-        if is_step_function_event(dd_context):
-            return extract_context_from_step_functions(dd_context, lambda_context)
+        try:
+            return extract_context_from_step_functions(dd_context, None)
+        except Exception:
+            logger.debug(
+                "Failed to extract Step Functions context from EventBridge event."
+            )
 
         return propagator.extract(dd_context)
     except Exception as e:
@@ -1322,13 +1326,16 @@ def create_inferred_span_from_eventbridge_event(event, context):
         tag_source="self",
     )
 
+    timestamp = event.get("time")
+    dt_format = "%Y-%m-%dT%H:%M:%SZ"
+
     # Use more granular timestamp from upstream Step Function if possible
-    if is_step_function_event(event.get("detail")):
-        timestamp = event.get("detail").get("_datadog").get("State").get("EnteredTime")
-        dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
-    else:
-        timestamp = event.get("time")
-        dt_format = "%Y-%m-%dT%H:%M:%SZ"
+    try:
+        if is_step_function_event(event.get("detail")):
+            timestamp = event["detail"]["_datadog"]["State"]["EnteredTime"]
+            dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
+    except (TypeError, KeyError, AttributeError):
+        logger.debug("Error parsing timestamp from Step Functions event")
 
     dt = datetime.strptime(timestamp, dt_format)
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -314,6 +314,14 @@ def _extract_context_from_eventbridge_sqs_event(event):
     body = json.loads(body_str)
     detail = body.get("detail")
     dd_context = detail.get("_datadog")
+    if is_step_function_event(dd_context):
+        try:
+            return extract_context_from_step_functions(detail, None)
+        except Exception:
+            logger.debug(
+                "Failed to extract Step Functions context from EventBridge to SQS event."
+            )
+
     return propagator.extract(dd_context)
 
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -39,6 +39,7 @@ from datadog_lambda.trigger import (
     _EventSource,
     parse_event_source,
     get_first_record,
+    is_step_function_event,
     EventTypes,
     EventSubtypes,
 )
@@ -477,32 +478,6 @@ def extract_context_from_step_functions(event, lambda_context):
     except Exception as e:
         logger.debug("The Step Functions trace extractor returned with error %s", e)
         return extract_context_from_lambda_context(lambda_context)
-
-
-def is_step_function_event(event):
-    """
-    Check if the event is a step function that invoked the current lambda.
-
-    The whole event can be wrapped in "Payload" in Legacy Lambda cases. There may also be a
-    "_datadog" for JSONata style context propagation.
-
-    The actual event must contain "Execution", "StateMachine", and "State" fields.
-    """
-    event = event.get("Payload", event)
-
-    # JSONPath style
-    if all(field in event for field in ("Execution", "StateMachine", "State")):
-        return True
-
-    # JSONata style
-    if "_datadog" in event:
-        event = event["_datadog"]
-        return all(
-            field in event
-            for field in ("Execution", "StateMachine", "State", "serverless-version")
-        )
-
-    return False
 
 
 def extract_context_custom_extractor(extractor, event, lambda_context):
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -1303,8 +1303,15 @@ def create_inferred_span_from_eventbridge_event(event, context):
         synchronicity="async",
         tag_source="self",
     )
-    dt_format = "%Y-%m-%dT%H:%M:%SZ"
-    timestamp = event.get("time")
+
+    # Use more granular timestamp from upstream Step Function if possible
+    if is_step_function_event(event.get("detail")):
+        timestamp = event.get("detail").get("_datadog").get("State").get("EnteredTime")
+        dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
+    else:
+        timestamp = event.get("time")
+        dt_format = "%Y-%m-%dT%H:%M:%SZ"
+
     dt = datetime.strptime(timestamp, dt_format)
 
     tracer.set_tags(_dd_origin)
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -272,6 +272,15 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
             if dd_json_data:
                 dd_data = json.loads(dd_json_data)
+
+                if is_step_function_event(dd_data):
+                    try:
+                        return extract_context_from_step_functions(dd_data, None)
+                    except Exception:
+                        logger.debug(
+                            "Failed to extract Step Functions context from SQS/SNS event."
+                        )
+
                 return propagator.extract(dd_data)
         else:
             # Handle case where trace context is injected into attributes.AWSTraceHeader
@@ -314,9 +323,10 @@ def _extract_context_from_eventbridge_sqs_event(event):
     body = json.loads(body_str)
     detail = body.get("detail")
     dd_context = detail.get("_datadog")
+
     if is_step_function_event(dd_context):
         try:
-            return extract_context_from_step_functions(detail, None)
+            return extract_context_from_step_functions(dd_context, None)
         except Exception:
             logger.debug(
                 "Failed to extract Step Functions context from EventBridge to SQS event."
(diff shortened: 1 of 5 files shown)


datadog_lambda/wrapper.py
```diff
@@ -279,8 +278,6 @@ def _before(self, event, context):
             self.response = None
             set_cold_start(init_timestamp_ns)
             submit_invocations_metric(context)
-            if is_legacy_lambda_step_function(event):
-                event = event["Payload"]
```
avedmala, line 283: Moved this unwrapping to happen inside of `tracing.extract_context_from_step_functions()`

datadog_lambda/tracing.py
```diff
@@ -1320,6 +1314,10 @@ def create_inferred_span_from_eventbridge_event(event, context):
     if span:
         span.set_tags(tags)
     span.start = dt.replace(tzinfo=timezone.utc).timestamp()
+
+    # Since inferred span will later parent Lambda, preserve Lambda's current parent
+    span.parent_id = dd_trace_context.span_id
```
avedmala, line 1319: This is important because we have the following code in `tracing.create_function_execution_span()`
```
if parent_span:
    span.parent_id = parent_span.span_id
```
where `parent_span` is the generated inferred span so the Lambda's root span's `parent_id` will be set to the inferred span's `span_id`

If there is an upstream Step Function and we saved its trace context in `dd_trace_context`, we want to preserve the parenting relationship and not let the inferred span completely erase it

This line solves the issue by making the inferred span be a child of the upstream service

datadog_lambda/tracing.py
```diff
@@ -1309,8 +1303,15 @@ def create_inferred_span_from_eventbridge_event(event, context):
         synchronicity="async",
         tag_source="self",
     )
-    dt_format = "%Y-%m-%dT%H:%M:%SZ"
-    timestamp = event.get("time")
+
+    # Use more granular timestamp from upstream Step Function if possible
+    if is_step_function_event(event.get("detail")):
+        timestamp = event.get("detail").get("_datadog").get("State").get("EnteredTime")
+        dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
```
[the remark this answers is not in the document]
kimi-p: Great job!

ps. I was confused about what `(1)` and `(2)` mean when I first glanced the PR. Originally, I thought that's some logic before this section of code that got hidden on the github.

datadog_lambda/tracing.py
```diff
@@ -1320,6 +1329,10 @@ def create_inferred_span_from_eventbridge_event(event, context):
     if span:
         span.set_tags(tags)
     span.start = dt.replace(tzinfo=timezone.utc).timestamp()
+
+    # Since inferred span will later parent Lambda, preserve Lambda's current parent
+    span.parent_id = dd_trace_context.span_id
```
kimi-p, line 1334: It seems that `dd_trace_context` would always have span_id. But, in case there is a bug or some edge cases, we may want to do type check and field exist check especially in python where event typing is not strongly enforced.

datadog_lambda/tracing.py
```diff
@@ -313,19 +314,32 @@ def _extract_context_from_eventbridge_sqs_event(event):
     body = json.loads(body_str)
     detail = body.get("detail")
     dd_context = detail.get("_datadog")
+    if is_step_function_event(dd_context):
+        try:
+            return extract_context_from_step_functions(detail, None)
```
kimi-p, line 319: If SFN -> EventBridge -> SQS -> Lambda has a different input format, can you add a test case? Thanks!

kimi-p (approved):
Only minor nix.  Really appreciate these great in-depth comments! Thank you!

datadog_lambda/tracing.py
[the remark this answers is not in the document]
avedmala: Added numbers next to the pictures to make it clearer for other reviewers

datadog_lambda/trigger.py
```diff
@@ -369,3 +367,28 @@ def extract_http_status_code_tag(trigger_tags, response):
         status_code = response.status_code
 
     return str(status_code)
+
+
+def is_step_function_event(event):
```
purple4reina, line 372: Is there a way we can memoize this function? It looks like it can potentially be called several times in the course of a single invocation.

purple4reina: Hmm, or it looks like the function can be called multiple times per invocation, but with different "events" each time? If that's true, then we can probably leave it.

datadog_lambda/trigger.py
```diff
@@ -369,3 +367,28 @@ def extract_http_status_code_tag(trigger_tags, response):
...
+
+    The whole event can be wrapped in "Payload" in Legacy Lambda cases. There may also be a
+    "_datadog" for JSONata style context propagation.
+
+    The actual event must contain "Execution", "StateMachine", and "State" fields.
+    """
+    event = event.get("Payload", event)
+
+    # JSONPath style
+    if all(field in event for field in ("Execution", "StateMachine", "State")):
```
purple4reina, line 384: I hate to say it (cuz this code is so easy to read and maintain as it is) that using `all` and the tuple are probably slower than just doing it manually. For the way it's written now, you gotta allocate the tuple, plus create the generator, all before even doing any work to evaluate the `all`.

I'd be curious to know the timing difference though. I'd recommend using the `timeit` module, or just running some benchmarks manually.

Here's what I'm thinking for the naive approach to compare this to:

```py
if "Execution" in event and "StateMachine" in event and "State" in event:
    return True
```

datadog_lambda/tracing.py
```diff
@@ -313,19 +323,35 @@ def _extract_context_from_eventbridge_sqs_event(event):
...
+    that header.
     """
     try:
         detail = event.get("detail")
         dd_context = detail.get("_datadog")
         if not dd_context:
             return extract_context_from_lambda_context(lambda_context)
+
+        if is_step_function_event(dd_context):
+            return extract_context_from_step_functions(dd_context, lambda_context)
```
purple4reina, line 353: This one isn't wrapped in a try/except, but the two above are, why is that?

datadog_lambda/tracing.py
```diff
@@ -1309,8 +1321,15 @@ def create_inferred_span_from_eventbridge_event(event, context):
         synchronicity="async",
         tag_source="self",
     )
-    dt_format = "%Y-%m-%dT%H:%M:%SZ"
-    timestamp = event.get("time")
+
+    # Use more granular timestamp from upstream Step Function if possible
+    if is_step_function_event(event.get("detail")):
+        timestamp = event.get("detail").get("_datadog").get("State").get("EnteredTime")
```
purple4reina, line 1327: Is it possible that these keys are not in the event? If so, this will raise an exception which I'm not sure is being caught?

purple4reina (commented):
Left some questions about performance and exception handling.

datadog_lambda/tracing.py
avedmala: Ah I also meant to wrap this in try/except but lemme explain why I'm doing it

I wanted to pass in `lambda_context=None` so that if the extractor fails, it won't fallback on the `lambda_context` and it'll continue with the normal codepath and call
```
return propagator.extract(dd_context)
```

datadog_lambda/tracing.py
avedmala: It could be possible if a customer messes up the instrumentation

I think it's worth handling tho so lemme add some logic to fallback to `event.get("time")` if this fails

datadog_lambda/trigger.py
avedmala: Did a quick benchmark and looks like using `all()` is >4x slower 😮

I'll switch it back to simple approach

datadog_lambda/trigger.py
avedmala: That's a great idea!

Correct me if I'm wrong but does the layer only handle one `event` per invocation? Or if it's a busy Lambda does it stay alive and potentially handle hundreds of events?

Just wondering to get an idea of how large to make the cache. I guess it can be pretty small anyway since each event is new and we don't repeat

purple4reina: Each runtime instance will only ever handle one event at a time. It never handles two events concurrently.

avedmala: Ah just realized we can't memoize it because `event` is a `dict` and mutable types are unhashable

We could serialize the dict and use that but I'm thinking that'd be much slower

datadog_lambda/tracing.py
```diff
@@ -313,19 +323,39 @@ def _extract_context_from_eventbridge_sqs_event(event):
...
+            return extract_context_from_step_functions(dd_context, None)
+        except Exception:
+            logger.debug(
+                "Failed to extract Step Functions context from EventBridge to SQS event."
+            )
+
     return propagator.extract(dd_context)
 
 
 def extract_context_from_eventbridge_event(event, lambda_context):
```
nine5two7, line 338: I am curious about how the concatenation of two queues (e.g., SFN → EventBridge → SQS → Lambda) is handled. Is it achieved by extracting two different contexts in the Python tracer? Does this mean that it also supports SFN → EventBridge → SQS → SNS → Lambda?

avedmala: SFN → EventBridge → SQS → Lambda is handled the following way
* Final event that enters the Lambda is an SQS event
* Our context extractor for SQS events checks if there's an EventBridge event within and uses that if its valid

SFN → SNS → SQS → Lambda is handled very similarly with another explicitly check in the SQS extractor looking for SNS events nested

We don't handle SFN → SQS → SNS → Lambda AFAIK but we wouldn't be able to handle SFN → EventBridge → SQS → SNS → Lambda out of the box either

But this is only because it's not explicitly handled. The current python layer implementation is messy because it relies on explicit handling. I think a perfect solution would be one where it's all handled recursively and customers can nest an arbitrary number of supported services without explicit handling

I think AWS team would like to do something like this in bottlecap

nine5two7: @avedmala Thanks for the explanation. Very informative.

purple4reina (approved):

tests/test_tracing.py
```diff
@@ -812,28 +759,183 @@ def test_step_function_trace_data_sfn_root(self):
...
+        )
+
+    @with_trace_propagation_style("datadog")
+    def test_step_function_trace_data_sns(self):
+        """Test step function trace data extraction through SNS"""
+        sns_event = {
+            "Records": [
+                {
+                    "EventSource": "aws:sns",
+                    "EventVersion": "1.0",
```
nhulston, line 887: Wow this file is getting really long! Not your job, but a future task for the serverless AWS team could be to pull our these events as JSON files and import the json files in the tests, instead of defining the objects directly here

nhulston (approved):
Looks good to me! Super thorough, and nice job on manually testing + sharing screenshots for trace propagation case

datadog_lambda/tracing.py
joeyzhao2018: Regarding the "recursive solution", is it written down in any RFC? it sounds interesting and might be able to solve some other problems.

PR #581, opened 2025-03-28
chore: lazy load some imports

duncanista:
### What does this PR do?

Lazy loads some imports for metrics and hashlib

### Motivation

Metrics submission for invocations and errors doesn't need to be called if there's an extension active.
Hashlib import shouldn't occur unless the method for step functions is called.

### Testing Guidelines

Tested manually

Before:
Screenshot 2025-03-28 at 4 25 43 PM

After:
Screenshot 2025-03-28 at 4 24 18 PM

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -23,11 +23,6 @@ from datadog_lambda.constants import (
     XraySubsegment,
     Headers,
 )
-from datadog_lambda.metric import (
-    flush_stats,
-    submit_invocations_metric,
-    submit_errors_metric,
-)
 from datadog_lambda.module_name import modify_module_name
 from datadog_lambda.patch import patch_all
 from datadog_lambda.span_pointers import calculate_span_pointers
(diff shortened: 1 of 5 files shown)


datadog_lambda/wrapper.py
```diff
@@ -248,7 +243,11 @@ def __call__(self, event, context, **kwargs):
             self.response = self.func(event, context, **kwargs)
             return self.response
         except Exception:
-            submit_errors_metric(context)
+            if not should_use_extension:
+                from datadog_lambda.metric import submit_invocations_metric
+
+                submit_invocations_metric(context)
```
purple4reina, line 249: I don't think we want this change. You replaced an errors metric with an invocation metric. 😄

purple4reina (changes requested):
The errors metric was changed into an invocation metric.

datadog_lambda/wrapper.py
duncanista: Woops, thanks for catching that!

purple4reina (approved):

nhulston (approved):

PR #584, opened 2025-04-10
Run e2e tests on each PR.

purple4reina:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


ci/input_files/build.yaml.tpl
```diff
@@ -3,6 +3,7 @@ stages:
  - test
  - sign
  - publish
+ - e2e
```
purple4reina, line 8: `e2e` needs `publish` which needs `test`

ci/input_files/build.yaml.tpl
```diff
@@ -132,16 +135,22 @@ sign-layer ({{ $runtime.name }}-{{ $runtime.arch }}):
...
 {{ range $environment_name, $environment := (ds "environments").environments }}
+{{ $dotenv := print $runtime.name "_" $runtime.arch "_" $environment_name ".env" }}
 
 publish-layer-{{ $environment_name }} ({{ $runtime.name }}-{{ $runtime.arch }}):
   stage: publish
   tags: ["arch:amd64"]
   image: registry.ddbuild.io/images/docker:20.10-py3
   rules:
+    - if: '"{{ $environment_name }}" == "sandbox" && $REGION == "{{ $e2e_region }}" && "{{ $runtime.arch }}" == "amd64"'
+      when: on_success
```
purple4reina, line 146: Always be sure to publish to the sandbox account in us-west-2 for e2e testing.

ci/input_files/build.yaml.tpl
```diff
@@ -132,16 +135,22 @@ sign-layer ({{ $runtime.name }}-{{ $runtime.arch }}):
...
     - if: '"{{ $environment_name }}" == "sandbox"'
       when: manual
       allow_failure: true
     - if: '$CI_COMMIT_TAG =~ /^v.*/'
+  artifacts:
+    reports:
+      dotenv: {{ $dotenv }}
```
purple4reina, line 153: When saved as a dotenv report, this file will get sourced for its env vars in every downstream job.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,53 @@ signed layer bundle:
...
+  trigger:
+    project: DataDog/serverless-e2e-tests
+    strategy: depend
+  variables:
+      LANGUAGES_SUBSET: python
+      # These env vars are inherited from the dotenv reports of the publish-layer jobs
+    {{- range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+    {{- $version := print (.name | strings.Trim "python") }}
+      PYTHON_{{ $version }}_VERSION: $PYTHON_{{ $version }}_VERSION
```
purple4reina, line 256: Explicitly pass environment variables from the sourced publish job dotenv file to the downstream job. These env vars are not passed downstream otherwise.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,53 @@ signed layer bundle:
...
+    {{- end }}
+    {{- end }}
+  needs: {{ range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+      - "publish-layer-sandbox ({{ .name }}-{{ .arch }}): [{{ $e2e_region }}]"
```
purple4reina, line 261: Run after all the publish jobs have completed.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,53 @@ signed layer bundle:
...
+    {{- end }}
+    {{- end }}
+
+e2e-status:
```
purple4reina: The purpose of this job is to simply relay the status of the downstream job back to the github pull request. Without this, the downstream e2e tests still run, and the `e2e-test` job will still properly fail. However, none of those failures would be posted to github, because gitlab will only post direct jobs, not jobs that it's triggering in another repo.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,53 @@ signed layer bundle:
...
+  stage: e2e
+  image: registry.ddbuild.io/images/mirror/alpine:latest
+  tags: ["arch:amd64"]
+  needs:
+    - e2e-test
+    {{- range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+    - "publish-layer-sandbox ({{ .name }}-{{ .arch }}): [{{ $e2e_region }}]"
```
purple4reina: We need to need these in order to get the `PYTHON_x_VERSION` env vars.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,53 @@ signed layer bundle:
...
+      - "publish-layer-sandbox ({{ .name }}-{{ .arch }}): [{{ $e2e_region }}]"
+    {{- end }}
+    {{- end }}
+
+e2e-status:
+  stage: e2e
+  image: registry.ddbuild.io/images/mirror/alpine:latest
+  tags: ["arch:amd64"]
+  needs:
+    - e2e-test
```
purple4reina: This job will run once the `e2e-test` job completes.

ci/publish_layers.sh
```diff
@@ -199,6 +199,8 @@ fi
 while [ $latest_version -lt $VERSION ]; do
     latest_version=$(publish_layer $REGION $layer $aws_cli_python_version_key $layer_path)
     printf "[$REGION] Published version $latest_version for layer $layer in region $REGION\n"
+    latest_arn=$(aws lambda get-layer-version --layer-name $layer --version-number $latest_version --region $REGION --query 'LayerVersionArn' --output text)
+    printf "[$REGION] Published arn $latest_arn\n"
```
purple4reina, line 203: Be sur to always print the arn to stdout when publishing layers. The publish gitlab jobs will do a regex search on this stdout to determine the arn for a newly published layer.

ci/input_files/build.yaml.tpl
purple4reina: On second thought, I don't actually think these are required. But I've left them in anyway just in case.

ci/input_files/build.yaml.tpl
```diff
@@ -1,8 +1,11 @@
+{{- $e2e_region := "us-west-2" -}}
```
duncanista, line 1: Is this a variable declaration? If so, why not use an environment variable?

ci/input_files/build.yaml.tpl
```diff
@@ -166,7 +175,19 @@ publish-layer-{{ $environment_name }} ({{ $runtime.name }}-{{ $runtime.arch }}):
...
+      # Extract the arn from the publish log to be used as envvar in e2e tests
+      layer_arn="$(grep 'Published arn' publish.log | grep -oE 'arn:aws:lambda:.*')"
+      if [ -z "$layer_arn" ]; then
+        echo "Error: Layer ARN not found in publish log"
+        exit 1
+      else
+        echo "Found layer arn, $layer_arn"
+      fi
+      echo "PYTHON_{{ $runtime.name | strings.Trim "python" }}_VERSION=$layer_arn" > {{ $dotenv }}
+      cat {{ $dotenv }}
```
duncanista: It would be ideal to not have scripts in the template definition, whereas we just call bash scripts

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,56 @@ signed layer bundle:
...
+    {{- end }}
+  needs: {{ range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+      - "publish-layer-sandbox ({{ .name }}-{{ .arch }}): [{{ $e2e_region }}]"
+    {{- end }}
+    {{- end }}
+
+e2e-status:
+  stage: e2e
+  image: registry.ddbuild.io/images/mirror/alpine:latest
```
duncanista: why do you need an alpine image here?

ci/input_files/build.yaml.tpl
duncanista: I don't think you need this

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +253,56 @@ signed layer bundle:
...
+  script:
+    - echo "Python layer ARNs used in E2E tests:"
+    {{- range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+    {{- $version := print (.name | strings.Trim "python") }}
+    - echo "  PYTHON_{{ $version }}_VERSION=$PYTHON_{{ $version }}_VERSION"
+    {{- end }}
+    {{- end }}
+    - |
+      if [ "$CI_JOB_STATUS" = "success" ]; then
```
duncanista: According to the docs, this is for the current executing job.

I'd suggest looking at how we check another's job status using the GitLab API

https://github.com/DataDog/dd-trace-js/blob/b269a7265f46955ab447a2b5a705cc08c4a959fb/.gitlab/benchmarks.yml#L89-L98

ci/input_files/build.yaml.tpl
purple4reina: Yes, it is just a variable declaration. I hadn't thought of just using an environment variable. I was thinking that this way, the value is just built into the final yaml file, instead of being resolved later, at ci execution time. What are your thoughts on that?

Ah, okay, so one of the places this variable is used is in a `needs` block. According to [gitlab docs](https://docs.gitlab.com/ci/variables/where_variables_can_be_used/), variables are not available in `needs` blocks.

ci/publish_layers.sh
```diff
@@ -209,4 +211,10 @@ while [ $latest_version -lt $VERSION ]; do
     fi
 done
 
+if [ -n "$DOTENV" ]; then
+    printf "[$REGION] Exporting layer version to $DOTENV file...\n"
+    echo "PYTHON_${PYTHON_VERSION/./}_VERSION=$latest_arn" >> "$DOTENV"
+    cat "$DOTENV"
```
purple4reina, line 217: Here is where we write the arn to the given dotenv file. This dotenv file will be saved as an artifact on the job.

ci/input_files/build.yaml.tpl
```diff
@@ -232,3 +241,47 @@ signed layer bundle:
...
+      PYTHON_{{ $version }}_VERSION: $PYTHON_{{ $version }}_VERSION
+    {{- end }}
+    {{- end }}
+  needs: {{ range (ds "runtimes").runtimes }}
+    {{- if eq .arch "amd64" }}
+      - "publish-layer-sandbox ({{ .name }}-{{ .arch }}): [{{ $e2e_region }}]"
+    {{- end }}
+    {{- end }}
+
+e2e-test-status:
```
purple4reina, line 265: This `e2e-test-status` job will be overhauled in a later PR.

ci/publish_layers.sh
purple4reina: Update, the publish gitlab job will _not_ be doing a regex search. Instead we will write the layer versions to a dotenv file, as seen below.

happynancee (approved):

PR #588, opened 2025-05-02
feat: Correct FIPS-mode metrics

apiarian-datadog:
- Our `dogstatsd` client now supports timestamps for the metrics that it will send.
- This unblocks us to always send metrics to the extension, even if they have a timestamp. Confirmed that this actually works now with both bottlecap and the go agent.
- Refactored the metrics workflow to have an explicit choice of metrics handlers (Extension, Forwarder, Datadog API, or, for some FIPS usecases, No Handler).
- Added a `DD_LAMBDA_FIPS_MODE` flag which allows FIPS-mode logic to be enabled in commercial regions or disabled in govcloud regions.
- The new FIPS mode is used for Datadog API Key secret lookup and for metrics handling decisions.

### Breaking Change
Since the `DD_LAMBDA_FIPS_MODE` defaults to `true` in govcloud, direct metrics submission there (without an Extension or a Forwarder) will now be disabled.

### Testing Guidelines
Unit tests were added or updated. Also confirmed with test apps that this logic works as expected.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/fips.py
+++ b/datadog_lambda/fips.py
@@ -1,0 +1,19 @@
+import logging
+import os
+
+is_gov_region = os.environ.get("AWS_REGION", "").startswith("us-gov-")
+
+enable_fips_mode = (
+    os.environ.get(
+        "DD_LAMBDA_FIPS_MODE",
+        "true" if is_gov_region else "false",
+    ).lower()
+    == "true"
+)
+
+if is_gov_region or enable_fips_mode:
+    logger = logging.getLogger(__name__)
+    logger.debug(
+        "Python Lambda Layer FIPS mode is %s.",
+        "enabled" if enable_fips_mode else "not enabled",
+    )
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -12,6 +12,7 @@ from datetime import datetime, timedelta
 import ujson as json
 
 from datadog_lambda.extension import should_use_extension
+from datadog_lambda.fips import enable_fips_mode
 from datadog_lambda.tags import dd_lambda_layer_tag, get_enhanced_metrics_tags
 
 logger = logging.getLogger(__name__)
@@ -21,6 +22,7 @@ class MetricsHandler(enum.Enum):
     EXTENSION = "extension"
     FORWARDER = "forwarder"
     DATADOG_API = "datadog_api"
+    NO_METRICS = "no_metrics"
 
 
 def _select_metrics_handler():
@@ -28,10 +30,19 @@ def _select_metrics_handler():
         return MetricsHandler.EXTENSION
     if os.environ.get("DD_FLUSH_TO_LOG", "").lower() == "true":
         return MetricsHandler.FORWARDER
+
+    if enable_fips_mode:
+        logger.debug(
+            "With FIPS mode enabled, the Datadog API metrics handler is unavailable."
+        )
+        return MetricsHandler.NO_METRICS
+
     return MetricsHandler.DATADOG_API
 
 
 metrics_handler = _select_metrics_handler()
+# TODO: add a metric for this so that we can see how often the DATADOG_API
+# metrics handler is actually used.
 logger.debug("identified primary metrics handler as %s", metrics_handler)
 
 
@@ -121,6 +132,12 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
 
     elif metrics_handler == MetricsHandler.DATADOG_API:
         lambda_stats.distribution(metric_name, value, tags=tags, timestamp=timestamp)
+
+    elif metrics_handler == MetricsHandler.NO_METRICS:
+        logger.debug(
+            "Metric %s cannot be submitted because the metrics handler is disabled: %s",
+            metric_name,
+        ),
 
     else:
         # This should be qutie impossible, but let's at least log a message if
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -41,8 +41,6 @@ def _select_metrics_handler():
 
 
 metrics_handler = _select_metrics_handler()
-# TODO: add a metric for this so that we can see how often the DATADOG_API
-# metrics handler is actually used.
 logger.debug("identified primary metrics handler as %s", metrics_handler)
 
 
--- a/datadog_lambda/metric.py
+++ b/datadog_lambda/metric.py
@@ -76,39 +76,25 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
     tags = [] if tags is None else list(tags)
     tags.append(dd_lambda_layer_tag)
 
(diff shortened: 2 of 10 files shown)


apiarian-datadog:
Probably easiest to review this one commit at a time.

datadog_lambda/metric.py
```diff
@@ -3,37 +3,66 @@
...
+        logger.debug(
+            "With FIPS mode enabled, the Datadog API metrics handler is unavailable."
+        )
+        return MetricsHandler.NO_METRICS
+
+    return MetricsHandler.DATADOG_API
+
+
+metrics_handler = _select_metrics_handler()
+logger.debug("identified primary metrics handler as %s", metrics_handler)
```
apiarian-datadog, line 44: it would be nice to fire off a metric here, similar to the way we do for dynamodb stream settings. but since  the most interesting value, `NO_METRICS` would actually be unavailable, i chose not to do this. let me know if you think we should still send the metric anyway, despite the `NO_METRICS` blind spot.

datadog_lambda/metric.py
```diff
@@ -71,56 +103,54 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
...
         logger.debug(
             "Sending metric %s value %s to Datadog via extension", metric_name, value
         )
         lambda_stats.distribution(metric_name, value, tags=tags, timestamp=timestamp)
+
+    elif force_async or (metrics_handler == MetricsHandler.FORWARDER):
+        write_metric_point_to_stdout(metric_name, value, timestamp=timestamp, tags=tags)
+
+    elif metrics_handler == MetricsHandler.DATADOG_API:
+        lambda_stats.distribution(metric_name, value, tags=tags, timestamp=timestamp)
```
apiarian-datadog, line 132: all of the other metric calls log a debug message when they're sending the metric. should we add one here?

apiarian-datadog:
confirmed that this is working as expected across the following combinations:
- bottlecap
- bottlecap fipsish
- go extension
- go extension fipsish
- forwarder
- forwarder fips-ish
- no extension and no forwarder

(all of the above correctly send metrics with and without timestamps)

the no extension and no forwarder case in fipsish mode does not send metrics as expected.

datadog_lambda/metric.py
```diff
@@ -71,56 +103,54 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
...
+                timestamp = int(timestamp.timestamp())
+
+            timestamp_floor = int((datetime.now() - timedelta(hours=4)).timestamp())
+            if timestamp < timestamp_floor:
+                logger.warning(
+                    "Timestamp %s is older than 4 hours, not submitting metric %s",
+                    timestamp,
+                    metric_name,
+                )
+                return
```
purple4reina, line 121: Should we try to submit the metric anyway? Let the agent or intake be the ones to drop it, not us. This could be helpful should in the future these other things support >4hr backdates, we wouldn't have to make a change here.

datadog_lambda/metric.py
```diff
@@ -71,56 +103,54 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
...
     else:
-        if flush_to_logs or force_async:
-            write_metric_point_to_stdout(
-                metric_name, value, timestamp=timestamp, tags=tags
-            )
-        else:
-            lambda_stats.distribution(
-                metric_name, value, tags=tags, timestamp=timestamp
-            )
+        # This should be qutie impossible, but let's at least log a message if
```
purple4reina, line 141:
```suggestion
        # This should be quite impossible, but let's at least log a message if
```

datadog_lambda/metric.py
```diff
@@ -71,56 +103,54 @@ def lambda_metric(metric_name, value, timestamp=None, tags=None, force_async=Fal
...
+            "Metric %s cannot be submitted because the metrics handler is not configured: %s",
+            metric_name,
+            metrics_handler,
+        )
 
 
-def write_metric_point_to_stdout(metric_name, value, timestamp=None, tags=[]):
+def write_metric_point_to_stdout(metric_name, value, timestamp=None, tags=None):
     """Writes the specified metric point to standard output"""
+    tags = tags or []
```
purple4reina, line 152: Good catch!

purple4reina (approved):

datadog_lambda/metric.py
apiarian-datadog: there was a apparently some sort of issue with incorrect metrics in the case of >4 hours? i think @DylanLovesCoffee knows something about this? in either case i'd prefer to remove this restriction in a separate pr since this is already a pretty non-trivial change. https://datadoghq.atlassian.net/browse/SVLS-6766

datadog_lambda/fips.py
```diff
@@ -0,0 +1,19 @@
...
+    os.environ.get(
+        "DD_LAMBDA_FIPS_MODE",
+        "true" if is_gov_region else "false",
+    ).lower()
+    == "true"
+)
+
+if is_gov_region or enable_fips_mode:
+    logger = logging.getLogger(__name__)
+    logger.debug(
```
duncanista, line 16: We are logging on every cold start, just to update if this is expected

duncanista (approved):

datadog_lambda/fips.py
```diff
@@ -0,0 +1,19 @@
+import logging
+import os
+
+is_gov_region = os.environ.get("AWS_REGION", "").startswith("us-gov-")
+
+enable_fips_mode = (
```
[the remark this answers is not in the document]
apiarian-datadog: will do

datadog_lambda/fips.py
apiarian-datadog: yup, that's the plan so that we always log in gov and at least log in commercial fips

datadog_lambda/fips.py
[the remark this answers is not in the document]
apiarian-datadog: fixed.

datadog_lambda/metric.py
```diff
@@ -138,19 +168,8 @@ def write_metric_point_to_stdout(metric_name, value, timestamp=None, tags=[]):
 
 
 def flush_stats(lambda_context=None):
-    lambda_stats.flush()
-
-    if extension_thread_stats is not None:
-        tags = None
-        if lambda_context is not None:
-            tags = get_enhanced_metrics_tags(lambda_context)
```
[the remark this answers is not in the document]
apiarian-datadog: we still use it as part of the `submit_enhanced_metric` call. but the `extension_thread_stats` was never `None` with the changes i put in and thus this code was not being called anymore anyway. this get function have important side effects we need to bring back somewhere?

datadog_lambda/metric.py
apiarian-datadog: not for now.

PR #593, opened 2025-05-12
fix:  safely getting all the values for trigger tags

joeyzhao2018:
### What does this PR do?

Make sure we get the `trigger_tags` values safely without raising exceptions. We have some customer's payload that cannot be properly handled currently.

As an alternative, there's a `try-catch-call`[solution](https://github.com/DataDog/datadog-lambda-python/pull/595/files). (Not recommended)
The recommended approach is this PR.

### Motivation

https://datadoghq.atlassian.net/browse/APMSVLS-10
Customer ticket:
https://datadoghq.atlassian.net/browse/SLES-2218

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -174,6 +174,7 @@ def parse_event_source(event: dict) -> _EventSource:
 
 
 def detect_lambda_function_url_domain(domain: str) -> bool:
+    #  e.g. "etsn5fibjr.lambda-url.eu-south-1.amazonaws.com"
     if not isinstance(domain, str):
         return False
     domain_parts = domain.split(".")
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -114,10 +114,14 @@ def parse_event_source(event: dict) -> _EventSource:
 
     event_source = None
 
-    request_context = event.get("requestContext")
+    # Get requestContext safely and ensure it's a dictionary
+    request_context = event.get("requestContext", {})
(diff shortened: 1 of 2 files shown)


datadog_lambda/trigger.py
```diff
@@ -114,10 +114,14 @@ def parse_event_source(event: dict) -> _EventSource:
 
     event_source = None
 
-    request_context = event.get("requestContext")
+    # Get requestContext safely and ensure it's a dictionary
+    request_context = event.get("requestContext", {})
```
purple4reina, line 118: it's better to do `request_context = event.get("requestContext") or {}` becuase it will avoid an allocation in the case that the key is found in the dict.

datadog_lambda/trigger.py
```diff
@@ -283,17 +289,28 @@ def extract_http_tags(event):
     Extracts HTTP facet tags from the triggering event
     """
     http_tags = {}
-    request_context = event.get("requestContext")
+
+    # Safely get request_context and ensure it's a dictionary
+    request_context = event.get("requestContext", {})
+    if not isinstance(request_context, dict):
+        request_context = {}
```
purple4reina, line 296: Same here. I think we can do this more efficiently.

datadog_lambda/trigger.py
joeyzhao2018: that's some expert level advice on allocation, thanks!

PR #597, opened 2025-05-13
perf: fewer memory allocation

joeyzhao2018:
### What does this PR do?
Less memory allocation and better performance.

### Explanation
Memory allocation timing:

- d.get("k", {}) always creates a new empty dictionary when the function is called, regardless of whether it will be used
- d.get("k") or {} only creates the empty dictionary if d.get("k") returns a falsy value

Handling of falsy values:

- d.get("k", {}) only uses the default when the key is missing
- d.get("k") or {} uses the empty dict when the key is missing OR when the value exists but is falsy (None, empty string, 0, False, etc.)

Performance implications:

- d.get("k", {}) has a slight overhead of always creating an object that might be discarded
- d.get("k") or {} is slightly more efficient when the key exists with a truthy value

this is a follow-up PR for PR https://github.com/DataDog/datadog-lambda-python/pull/593.
https://datadoghq.atlassian.net/browse/APMSVLS-10

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -115,9 +115,9 @@ def parse_event_source(event: dict) -> _EventSource:
     event_source = None
 
     # Get requestContext safely and ensure it's a dictionary
-    request_context = event.get("requestContext", {})
+    request_context = event.get("requestContext") or {}
     if not isinstance(request_context, dict):
-        request_context = {}
+        request_context = None
 
     if request_context and request_context.get("stage"):
         if "domainName" in request_context and detect_lambda_function_url_domain(
@@ -291,9 +291,9 @@ def extract_http_tags(event):
     http_tags = {}
 
     # Safely get request_context and ensure it's a dictionary
-    request_context = event.get("requestContext", {})
+    request_context = event.get("requestContext") or {}
     if not isinstance(request_context, dict):
-        request_context = {}
+        request_context = None
 
     path = event.get("path")
     method = event.get("httpMethod")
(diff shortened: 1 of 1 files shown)


datadog_lambda/trigger.py
```diff
@@ -115,9 +115,9 @@ def parse_event_source(event: dict) -> _EventSource:
     event_source = None
 
     # Get requestContext safely and ensure it's a dictionary
-    request_context = event.get("requestContext", {})
+    request_context = event.get("requestContext") or {}
     if not isinstance(request_context, dict):
-        request_context = {}
+        request_context = None
```
purple4reina, line 120: Teeeeeechnically, in this situation you could just leave `event.get("requestContext")` without the `or {}`. Because you're properly handling the None case below.

datadog_lambda/trigger.py
```diff
@@ -115,9 +115,9 @@ def parse_event_source(event: dict) -> _EventSource:
     event_source = None
 
     # Get requestContext safely and ensure it's a dictionary
-    request_context = event.get("requestContext", {})
+    request_context = event.get("requestContext") or {}
```
joeyzhao2018, line 118:
```suggestion
    request_context = event.get("requestContext")
```

datadog_lambda/trigger.py
```diff
@@ -291,9 +291,9 @@ def extract_http_tags(event):
     http_tags = {}
 
     # Safely get request_context and ensure it's a dictionary
-    request_context = event.get("requestContext", {})
+    request_context = event.get("requestContext") or {}
```
joeyzhao2018, line 294:
```suggestion
    request_context = event.get("requestContext")
```

purple4reina (approved):

hghotra (approved):

PR #604, opened 2025-05-29
feat: Enable sqs -> lambda support for DSM

michael-zhao459:
### What does this PR do?

Allow DSM to support sqs -> lambda 

### Motivation

Lambda support requested by users of DSM

### Testing Guidelines

Wrote a unit test inside the test_wrapper.py code that ensures context is properly being set
throughout the entire pipeline

### Additional Notes

In the test I wrote a patched version of get_datastreams_context, see https://github.com/DataDog/dd-trace-py/pull/13526 for the change, forced to write patched function as no new release of tracer code yet. Will remove once there is a release of tracer code. 

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/dsm.py
+++ b/datadog_lambda/dsm.py
@@ -1,0 +1,37 @@
+from datadog_lambda import logger
+from datadog_lambda.trigger import EventTypes
+
+
+def set_dsm_context(event, event_source):
+
+    if event_source.equals(EventTypes.SQS):
+        _dsm_set_sqs_context(event)
+
+
+def _dsm_set_sqs_context(event):
+    from datadog_lambda.wrapper import format_err_with_traceback
+
+    from ddtrace.internal.datastreams.processor import (
+        DataStreamsProcessor as processor,
+        DsmPathwayCodec,
+    )
+    from ddtrace.internal.datastreams.botocore import (
+        get_datastreams_context,
+        calculate_sqs_payload_size,
+    )
+
+    records = event.get("Records", [])
+    for record in records:
+        try:
+            queue_arn = record.get("eventSourceARN", "")
+
+            contextjson = get_datastreams_context(record)
+            payload_size = calculate_sqs_payload_size(record)
+
+            ctx = DsmPathwayCodec.decode(contextjson, processor())
+            ctx.set_checkpoint(
+                ["direction:in", "topic:" + queue_arn, "type:sqs"],
+                payload_size=payload_size,
+            )
+        except Exception as e:
+            logger.error(format_err_with_traceback(e))
--- a/datadog_lambda/dsm.py
+++ b/datadog_lambda/dsm.py
@@ -10,11 +10,8 @@ def set_dsm_context(event, event_source):
 
 def _dsm_set_sqs_context(event):
     from datadog_lambda.wrapper import format_err_with_traceback
-
-    from ddtrace.internal.datastreams.processor import (
-        DataStreamsProcessor,
-        DsmPathwayCodec,
-    )
+    from ddtrace.internal.datastreams import data_streams_processor
+    from ddtrace.internal.datastreams.processor import DsmPathwayCodec
     from ddtrace.internal.datastreams.botocore import (
         get_datastreams_context,
         calculate_sqs_payload_size,
@@ -23,7 +20,7 @@ def _dsm_set_sqs_context(event):
     records = event.get("Records")
     if records is None:
         return
-    processor = DataStreamsProcessor()
+    processor = data_streams_processor()
 
     for record in records:
         try:
--- a/datadog_lambda/dsm.py
+++ b/datadog_lambda/dsm.py
@@ -12,7 +12,7 @@ def _dsm_set_sqs_context(event):
     from datadog_lambda.wrapper import format_err_with_traceback
 
     from ddtrace.internal.datastreams.processor import (
-        DataStreamsProcessor as processor,
+        DataStreamsProcessor,
         DsmPathwayCodec,
     )
     from ddtrace.internal.datastreams.botocore import (
@@ -20,7 +20,11 @@ def _dsm_set_sqs_context(event):
         calculate_sqs_payload_size,
     )
 
-    records = event.get("Records", [])
+    records = event.get("Records")
+    if records is None:
+        return
+    processor = DataStreamsProcessor()
+
     for record in records:
         try:
             queue_arn = record.get("eventSourceARN", "")
@@ -28,7 +32,7 @@ def _dsm_set_sqs_context(event):
             contextjson = get_datastreams_context(record)
             payload_size = calculate_sqs_payload_size(record)
 
-            ctx = DsmPathwayCodec.decode(contextjson, processor())
+            ctx = DsmPathwayCodec.decode(contextjson, processor)
             ctx.set_checkpoint(
                 ["direction:in", f"topic:{queue_arn}", "type:sqs"],
                 payload_size=payload_size,
--- a/datadog_lambda/dsm.py
+++ b/datadog_lambda/dsm.py
@@ -30,7 +30,7 @@ def _dsm_set_sqs_context(event):
 
             ctx = DsmPathwayCodec.decode(contextjson, processor())
             ctx.set_checkpoint(
-                ["direction:in", "topic:" + queue_arn, "type:sqs"],
+                ["direction:in", f"topic:{queue_arn}, "type:sqs"],
                 payload_size=payload_size,
             )
         except Exception as e:
--- a/datadog_lambda/dsm.py
+++ b/datadog_lambda/dsm.py
@@ -30,7 +30,7 @@ def _dsm_set_sqs_context(event):
 
             ctx = DsmPathwayCodec.decode(contextjson, processor())
             ctx.set_checkpoint(
-                ["direction:in", f"topic:{queue_arn}, "type:sqs"],
+                ["direction:in", f"topic:{queue_arn}", "type:sqs"],
                 payload_size=payload_size,
             )
         except Exception as e:
--- a/tests/test_dsm.py
+++ b/tests/test_dsm.py
@@ -1,0 +1,113 @@
+import unittest
+from unittest.mock import patch, MagicMock
+
+from datadog_lambda.dsm import set_dsm_context, _dsm_set_sqs_context
+from datadog_lambda.trigger import EventTypes
+
+
+class TestDsmContext(unittest.TestCase):
+    def test_non_sqs_event_source_does_nothing(self):
+        """Test that non-SQS event sources don't trigger DSM context setting"""
+        event = {"Records": [{"body": "test"}]}
+
+        mock_event_source = MagicMock()
+        mock_event_source.equals.return_value = False  # Not SQS
+
+        with patch("datadog_lambda.dsm._dsm_set_sqs_context") as mock_sqs_context:
+            set_dsm_context(event, mock_event_source)
+
+            mock_event_source.equals.assert_called_once_with(EventTypes.SQS)
+            mock_sqs_context.assert_not_called()
+
+    def test_event_with_no_records_does_nothing(self):
+        """Test that events where Records is None don't trigger DSM processing"""
+        events_with_no_records = [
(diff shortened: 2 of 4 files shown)


purple4reina (commented):
@michael-zhao459 the way your code works looks just fine. But as written right now, it's pretty inefficient, recalculates work already done, and needs some refactoring. Here's what I'm thinking.

We have a feature already implemented called "inferred spans", which similarly looks at the event type and does some stuff with the event payload based on the type. I'm thinking we'll want to create a mechanism similar to this.

Take a look in `datadog_lambda/tracing.py` at the `create_inferred_span` method and where it's called in `datadog_lambda/wrapper.py`. I'm thinking we'll want to add a method called `process_dsm` (or something like that) and call it from about the same spot we call `create_inferred_span`.

I can walk you through all of this during our pairing time.

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,37 @@
...
+
+    records = event.get("Records", [])
+    for record in records:
+        try:
+            queue_arn = record.get("eventSourceARN", "")
+
+            contextjson = get_datastreams_context(record)
+            payload_size = calculate_sqs_payload_size(record)
+
+            ctx = DsmPathwayCodec.decode(contextjson, processor())
```
purple4reina: I don't know enough about how this `DataStreamsProcessor` works. Is it intended to create a new instance of this type for each new record? Or is it okay for us to reuse the same processor over and over?

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,37 @@
...
+            ctx.set_checkpoint(
+                ["direction:in", "topic:" + queue_arn, "type:sqs"],
```
purple4reina: What we call "f strings" in python are more efficient than "+'ing" strings:

```suggestion
                ["direction:in", f"topic:{queue_arn}, "type:sqs"],
```

datadog_lambda/dsm.py
michael-zhao459: https://github.com/DataDog/dd-trace-py/blob/ac13894b14a6a668a01db9da8c65e58cb6162698/ddtrace/internal/datastreams/__init__.py#L19C1-L26C22, To my understanding, processor() is a singleton global instance that gets reused over and over

datadog_lambda/dsm.py
purple4reina: Also, one of the slowest things in python is the allocation of things like dictionaries and lists. You could make your code more efficient by reusing the same list over and over again. This of course requires special care that you're correctly updating the list on each loop, but I think in this case, the risk is pretty low. It does make the code a bit less straightforward to understand at a glance though.

Simplified, but here's what I'm getting at. Instead of

```python
for record in records:
    ...
    ctx.set_checkpoint(["direction:in", f"topic:{queue_arn}", "type:sqs"])
```

you could do something like

```python
tags = [None, "direction:in", "type:sqs"]
for record in records:
    ...
    tags[0] = f"topic:{queue_arn}"
    ctx.set_checkpoint(tags)
```

Important caveat: Since you'd be reusing the same tags object over and over again, you'd want to make sure that the `ctx.set_checkpoint` method doesn't use the original tags list, and instead is making a copy of it.

datadog_lambda/dsm.py
purple4reina: okay nice, then is it okay to use that global processor instead of creating a new one like you're doing every time here? Something like

```python
from ddtrace.internal.datastreams import data_streams_processor

processor = data_streams_processor()
for record in records:
    ...
    ctx = DsmPathwayCodec.decode(contextjson, processor())
```

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,37 @@
...
+    from ddtrace.internal.datastreams.processor import (
+        DataStreamsProcessor as processor,
+        DsmPathwayCodec,
+    )
+    from ddtrace.internal.datastreams.botocore import (
+        get_datastreams_context,
+        calculate_sqs_payload_size,
+    )
+
+    records = event.get("Records", [])
```
purple4reina: nit: even if a `"Records"` key is in the `event` dict, this will still allocate an empty list. Try this instead

```suggestion
    records = event.get("Records") or []
```

which only allocates the empty list when there are no records.

Though, even better yet, why allocate an empty list at all when you could just do

```python
records = event.get("Records")
if records is None:
    return
for record in records:
    ... do stuff ...
```

datadog_lambda/dsm.py
michael-zhao459: I think it should lemme test it real quick

tests/test_wrapper.py
```diff
@@ -563,6 +563,204 @@ def return_type_test(event, context):
             self.assertEqual(result, test_result)
             self.assertFalse(MockPrintExc.called)
 
+    @patch.dict(os.environ, {"DD_DATA_STREAMS_ENABLED": "true"})
+    def test_datadog_lambda_wrapper_dsm_sqs_context_pathway_verification(self):
```
purple4reina: I think we can vastly simplify these tests. This might warrant another pairing session, but I'll let you take a stab at it on your own first. Feel free to schedule something with me if you wanna go over it together.

There are two different files which you changed, and therefore two different test files that will need updating: `tests/test_wrapper.py` and a new file `tests/test_dsm.py`.

## test_wrapper.py

For the `test_wrapper.py` file, we simply need to test that, based on the env vars `DD_DATA_STREAMS_ENABLED` and `DD_TRACE_ENABLED` we either _do_ or _do not_ call `set_dsm_context` with the proper args. We'll push all of the verification of what happens inside of `set_dsm_context` to the `test_dsm.py` file.

I'm a bit fan of `pytest`, which isn't yet imported and used in this file, which allows you to reuse the same code over and over again to create "parametrized" tests. You can accomplish this same thing using `unittest` (as this file already uses), though it would mean creating 4 different test methods.

```python
# test_wrapper.py

import pytest

_test_set_dsm_context = (
		("true", "true", True),
		("true", "false", False),
		("false", "true", False),
		("false", "false", False),
)

@pytest.mark.parametrize("trace_enabled,dsm_enabled,should_call", _test_set_dsm_context)
def test_set_dsm_context(trace_enabled, dsm_enabled, should_call, monkeypatch):

	# use monkeypatch to set env vars DD_TRACE_ENABLED and DD_DATA_STREAMS_ENABLED
	
	# use monkeypatch to create a mock for `set_dsm_context`, you can also use mock.patch

	@wrapper.datadog_lambda_wrapper
	def lambda_handler(event, context):
		return "ok"

	result = lambda_handler(sqs_event, get_mock_context())
	assert result == "ok"

	if should_call: 
		# not sure of the api here, so this is just made up
		assert set_dsm_context_patch.called_with == (sqs_event, EventSource(EventSourceType.SQS))
	else:
		# again, not sure about api
		assert set_dsm_context.not_called
```

## test_dsm.py

In the `test_dsm.py` file, this is where you'll assert to make sure that the `set_dsm_context` works as expected. You'll want to include several tests:

1. Sending an event source of anything other than `SQS` will do nothing
2. Sending an event with no Records will do nothing
3. For each Record in the event, dsm does the setting of context as expected

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    from ddtrace.internal.datastreams.botocore import (
+        get_datastreams_context,
+        calculate_sqs_payload_size,
+    )
+
+    records = event.get("Records")
+    if records is None:
+        return
+    processor = DataStreamsProcessor()
```
purple4reina: According to my undersatnding of ddtrace (so please correct me if I'm wrong), initializing a new `DataStreamsProcessor` object starts a bunch of threads and has a whole ton of initialization code. We want to avoid doing that over and over again. Doing it like

```python
from ddtrace.internal.datastreams import data_streams_processor

processor = data_streams_processor()
```

on the other hand ensures that we use the same global processor throughout each invocation.

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def set_dsm_context(event, event_source):
+
+    if event_source.equals(EventTypes.SQS):
+        _dsm_set_sqs_context(event)
+
+
+def _dsm_set_sqs_context(event):
+    from datadog_lambda.wrapper import format_err_with_traceback
```
purple4reina, line 12: This import can be done at the top of the file

datadog_lambda/dsm.py
michael-zhao459: I get that its slower allocating a new list, but personally I think the current version is more readable and also I'm kinda afraid of reusing the same list for each set_checkpoint(), I would prefer doing what I am here

tests/test_wrapper.py
michael-zhao459: Lemme take a stab at these today, I will definitely schedule a meeting if I get stuck on any of these parts

datadog_lambda/dsm.py
michael-zhao459: I needed to move the import there to avoid circular imports, is there a better fix for this?

purple4reina: ah, I see, nah, at this point that's probably the best you can do.

tests/test_wrapper.py
```diff
@@ -563,6 +567,62 @@ def return_type_test(event, context):
...
+        @wrapper.datadog_lambda_wrapper
+        def lambda_handler(event, context):
+            return "ok"
+
+        result = lambda_handler({}, get_mock_context())
+        self.assertEqual(result, "ok")
+        self.mock_set_dsm_context.assert_not_called()
+
+        del os.environ["DD_DATA_STREAMS_ENABLED"]
+
```
purple4reina, line 625: These look great! Super easy to read and follow.

tests/test_dsm.py
```diff
@@ -0,0 +1,116 @@
...
+        self.mock_calculate_sqs_payload_size.return_value = 100
+        self.addCleanup(patcher.stop)
+
+        patcher = patch("ddtrace.internal.datastreams.processor.DsmPathwayCodec.decode")
+        self.mock_dsm_pathway_codec_decode = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_non_sqs_event_source_does_nothing(self):
+        """Test that non-SQS event sources don't trigger DSM context setting"""
+        event = {"Records": [{"body": "test"}]}
```
purple4reina: nit: Just to make this a tad less confusing, this event object sure looks like an sqs event to me. The test of course passes bc we look at the event source, not the event for its type. That said, can we make this event something like `{}`, just to make sure that it also _looks_ not sqs?

tests/test_dsm.py
```diff
@@ -0,0 +1,116 @@
...
+        """Test that events where Records is None don't trigger DSM processing"""
+        events_with_no_records = [
+            {},
+            {"Records": None},
+            {"someOtherField": "value"},
+        ]
+
+        for event in events_with_no_records:
+            _dsm_set_sqs_context(event)
+            self.mock_data_streams_processor.assert_not_called()
```
purple4reina, line 54: Very nice.

tests/test_dsm.py
```diff
@@ -0,0 +1,116 @@
...
+                {
+                    "eventSource": "aws:sqs",
+                    "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:my-queue",
+                    "body": "Hello from SQS!",
+                }
+            ]
+        }
+
+        mock_event_source = MagicMock()
+        mock_event_source.equals.return_value = True
```
purple4reina: I'm thinking toward the future, where you'll be adding support for sns and kinesis, etc. Once those are in place, you're not gonna want `event_source.equals(whatever)` to always return `True`.

Basically, this code works just great right now. But your future self is gonna have to change it, so why not make it work right the first time.

Instead, it's pretty easy to create a workable event source object. Instead of a mock, what about something like:

```python
event_source = _EventSource(EventTypes.SQS)

set_dsm_context(sqs_event, event_source)
```

Make that change to remove all the `mock_event_source` objects in all these tests.

PR #612, opened 2025-06-09
refactor: remove tracer dependencies to support dsm sqs -> lambda

michael-zhao459:
### What does this PR do?

Removes dependencies on internal tracer code. The get_dsm_context() is moved inside the lambda layer, and the public facing api for setting checkpoints is used instead of internal dsm code in the tracer.

### Motivation

Helps decouple the lambda layer code from the tracer code, keeps with bests practice of not using internal implementation code. 

### Testing Guidelines
Function was properly unit tested

### Additional Notes

IMPORTANT:  This PR cannot get merged until the tracer releases a version that includes this PR https://github.com/DataDog/dd-trace-py/pull/13646 where the manual_checkpoint parameter is added to the set_consume_checkpoint() code

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/tests/test_dsm.py
+++ b/tests/test_dsm.py
@@ -239,7 +239,7 @@ class TestSetDSMContext(unittest.TestCase):
 
         assert result is None
         self.mock_logger.debug.assert_called_with(
-            "DataStreams did not handle lambda message, context is not a dict, message: %r",
+            "DataStreams did not handle lambda message, dsm context is not a dict, message: %r",
             message_not_dict,
         )
 


michael-zhao459:
* **#612** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-lambda-python/612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-lambda-python/612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

datadog_lambda/dsm.py
```diff
@@ -24,15 +36,88 @@ def _dsm_set_sqs_context(event):
...
+        - message.messageAttributes._datadog.stringValue (SQS -> lambda)
+        - message.Sns.MessageAttributes._datadog.Value.decode() (SNS -> lambda)
+        - message.messageAttributes._datadog.binaryValue.decode() (SNS -> SQS -> lambda, raw)
+        - message.body.MessageAttributes._datadog.Value.decode() (SNS -> SQS -> lambda)
+        - message.kinesis.data.decode()._datadog (Kinesis -> lambda)
+    """
+    context_json = None
+    message_body = message
+
+    if "kinesis" in message:
```
purple4reina: Can we simplify this PR. Right now main doesn't even have kinesis support, so it's impossible for me to review this effectively. Start first with a single PR that does the change you're looking for but _just_ for SQS.

datadog_lambda/dsm.py
```diff
@@ -24,15 +36,88 @@ def _dsm_set_sqs_context(event):
...
+
+
+def _dsm_set_sqs_context(event):
+    from ddtrace.internal.datastreams.botocore import calculate_sqs_payload_size
+
+    def sqs_payload_calculator(record, context_json):
+        return calculate_sqs_payload_size(record)
+
+    def sqs_arn_extractor(record):
+        return record.get("eventSourceARN", "")
```
purple4reina: I'm not a fan of this approach. Why not just get the payload size and arn in this function and pass those values into the set helper?

michael-zhao459: When looping through each record they can definitely have different payload sizes, and when doing batch processing of records I'm pretty sure there could be differences in ARNs between the records. That's why I like having the for loop in the set_context_helper, to pass in payload size and arn to the set helper I would need that for loop inside of _dsm_set_sqs_context

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,79 @@
+import json
+
+from ddtrace.internal.logger import get_logger
```
purple4reina: This isn't in ddtrace so we shouldn't use the logger from it. Find a spot where we use a logger somewhere else in this repo and follow that.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,79 @@
...
+    processor = data_streams_processor()
+
+    try:
+        ctx = DsmPathwayCodec.decode(context_json, processor)
+        ctx.set_checkpoint(
+            ["direction:in", f"topic:{arn}", f"type:{service_type}"],
+            payload_size=payload_size,
+        )
+    except Exception as e:
+        logger.error(format_err_with_traceback(e))
```
purple4reina: I know this was here previously, but upon more thought, I think we shouldn't use `format_err_with_traceback`. Instead, we should just create a regular string message. The customer doesn't need the full traceback.

```suggestion
        logger.error(f"Unable to set dsm context: {e}")
```

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,61 @@
...
-            payload_size = calculate_sqs_payload_size(record)
-
-            ctx = DsmPathwayCodec.decode(contextjson, processor)
-            ctx.set_checkpoint(
-                ["direction:in", f"topic:{queue_arn}", "type:sqs"],
-                payload_size=payload_size,
-            )
-        except Exception as e:
-            logger.error(format_err_with_traceback(e))
+        arn = record.get("eventSourceARN", "")
```
piochelepiotr: we remove the try / except here. Is there a reason for that (maybe there is and I don't see it).
But we want to make sure our instrumentation never prevents the lambda from being executed, even if there is an issue with the instrumentation.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,61 @@
...
+        context_json = _get_dsm_context_from_lambda(record)
+        if not context_json:
+            logger.debug("DataStreams skipped lambda message: %r", record)
+            return None
```
piochelepiotr: just return is enough

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,61 @@
...
+
+        carrier_get = _create_carrier_get(context_json)
+        set_consume_checkpoint("sqs", arn, carrier_get)
```
piochelepiotr: (Let's make sure we update the set manual_checkpoint=False here once the other PR is merged & the tracer released.

michael-zhao459: Plan on keeping this PR in draft until tracer released, and will make sure to update

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,61 @@
...
+
+
+def _get_dsm_context_from_lambda(message):
+    """
+    Lambda-specific message formats:
+        - message.messageAttributes._datadog.stringValue (SQS -> lambda)
+    """
+    context_json = None
+    message_attributes = message.get("messageAttributes")
```
piochelepiotr, line 43: is it possible for the message attributes to be in the message body (for the sns --> sqs case for example)? Is that maybe in a separate PR?

michael-zhao459: This will be in separate PR. I am trying to keep this one solely based on fixing the sqs logic

tests/test_dsm.py
```diff
@@ -77,36 +68,152 @@ def test_sqs_multiple_records_process_each_record(self):
...
+        """Test message with MessageAttributes but no _datadog attribute returns None."""
+        message = {
+            "messageId": "test-message-id",
+            "body": "Test message",
+            "messageAttributes": {
+                "customAttribute": {"stringValue": "custom-value", "dataType": "String"}
+            },
+        }
+
+        result = _get_dsm_context_from_lambda(message)
```
robcarlan-datadog: [nit] newline here

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,68 @@
...
+    if not message_attributes:
+        logger.debug("DataStreams skipped lambda message: %r", message)
```
purple4reina: It looks like we're logging debug messages multiple times for the same record.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,68 @@
...
+        return None
+
+    if "_datadog" not in message_attributes:
+        logger.debug("DataStreams skipped lambda message: %r", message)
+        return None
+
+    datadog_attr = message_attributes["_datadog"]
+
+    if "stringValue" in datadog_attr:
```
purple4reina, line 66: We should do a type check here to ensure this is a dict.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,68 @@
...
+        # SQS -> lambda
```
purple4reina: We can use the `event_type` to avoid doing unnecessary work. We should already mostly know the shape of the event. Without doing so, this method is gonna get insanely large.

I would recommend creating a separate `_get_dsm_context` for each event type.

tests/test_dsm.py
```diff
@@ -77,36 +68,153 @@ def test_sqs_multiple_records_process_each_record(self):
...
+                            "dataType": "String",
+                        }
+                    },
                 },
                 {
                     "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:queue3",
                     "body": "Message 3",
+                    "messageAttributes": {
+                        "_datadog": {
+                            "stringValue": json.dumps(
```
purple4reina, line 102: We should add another test where the `stringValue` isn't a dict.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,68 @@
...
+
+
+def _set_dsm_context_for_record(record, type, arn):
+    from ddtrace.data_streams import set_consume_checkpoint
+
+    try:
+        context_json = _get_dsm_context_from_lambda(record)
+        if not context_json:
+            logger.debug("DataStreams skipped lambda message: %r", record)
+            return
```
purple4reina: This line is untested.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,68 @@
...
+
+        carrier_get = _create_carrier_get(context_json)
+        set_consume_checkpoint(type, arn, carrier_get, manual_checkpoint=False)
+    except Exception as e:
+        logger.error(f"Unable to set dsm context: {e}")
```
purple4reina: This line is untested. We should test to make sure that if there's an exception in setting the checkpoint that this method properly captures the error.

datadog_lambda/dsm.py
michael-zhao459: Just want to clarify but we are checking that context_json is a dict right? I think context_json is the only one we need to make sure is a dict, the test you asked me to write also signaled that to me

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,69 @@
...
     for record in records:
+        arn = record.get("eventSourceARN", "")
         try:
-            queue_arn = record.get("eventSourceARN", "")
-
-            contextjson = get_datastreams_context(record)
-            payload_size = calculate_sqs_payload_size(record)
+            context_json = _get_dsm_context_from_sqs_lambda(record)
+            if not context_json:
+                return
```
purple4reina: Should you continue instead of return?

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,69 @@
...
+
+    datadog_attr = message_attributes["_datadog"]
+
+    if "stringValue" in datadog_attr:
+        context_json = json.loads(datadog_attr["stringValue"])
+        if not isinstance(context_json, dict):
+            logger.debug("DataStreams did not handle lambda message: %r", message)
+            return None
+    else:
+        logger.debug("DataStreams did not handle lambda message: %r", message)
```
purple4reina: I would recommend making each of these log lines slightly different. That way when one is encountered, it is easy to find the exact line of code where it was produced. Otherwise, we don't know what the actual issue was.

datadog_lambda/dsm.py
purple4reina: Agreed, we should move this inside the try/except.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,78 @@
...
+            logger.debug(
+                "DataStreams did not handle lambda message: %r, context is not a dict",
+                message,
             )
-        except Exception as e:
-            logger.error(format_err_with_traceback(e))
+            return None
+    else:
+        logger.debug(
+            "DataStreams did not handle lambda message: %r, no dsm context", message
```
purple4reina: nit: put the `%r` at the end, the message itself could be quite long.

datadog_lambda/dsm.py
purple4reina: we should also make sure `datadog_attr` is a dict.

datadog_lambda/dsm.py
```diff
@@ -1,38 +1,87 @@
...
+            if not context_json:
+                continue
+            _set_dsm_context_for_record(context_json, "sqs", arn)
+
+        except Exception as e:
+            logger.error(f"Unable to set dsm context: {e}")
+
+
+def _set_dsm_context_for_record(context_json, type, arn):
+    from ddtrace.data_streams import set_consume_checkpoint
```
purple4reina, line 31: We need to be sure to set a minimum version for the ddtrace dependency. To do that, you'll want to find the first version of ddtrace that includes this `set_consume_checkpoint`. Then update pyproject.toml with this version.

michael-zhao459: Got it! Thanks for letting me know how this is done

purple4reina:
Is this PR still necessary? If not, it should be closed.

PR #621, opened 2025-06-19
feat: initial support for ASM inside the tracer

florentinl:
Adds **Application Security Management (ASM) – Threat Detection** support to the Python Lambda layer.

### What does this PR do?

* **Ship [`libddwaf.so`](https://github.com/DataDog/libddwaf) in the layer**

  * Shared library that powers the core security‑rules engine
  * Size: **2.6 MB** (adds ≈ 0.8 MB to the compressed layer)
  * *Note:* This requires to bump the layer size limit in the CI
* **Parse incoming HTTP events** and forward enriched data to ASM when enabled

  * Activation flag: `DD_APPSEC_ENABLED`
  * *Note:* This is deliberately distinct from `DD_SERVERLESS_APPSEC_ENABLED` (extension‑level enablement) currently in use by ASM clients in lambda

### Motivation

Enable AppSec through the tracer in the context of AWS Lambda

### Testing Guidelines

- unit tests for event information extraction
- benchmarking (using this [repo](https://github.com/purple4reina/dd-python-cold-start-testing/blob/main/dashboard.json)) with `DD_APPSEC_ENABLED=false` to ensure that there is no impact with ASM disabled. `before` is `main` and `after` is this branch both using `dd-trace-py` from main :
image

- manual testing for now for ASM, integration tests will be done later in [system-tests](https://github.com/DataDog/system-tests)

#### Additional Notes

With appsec enabled during the benchmarks, I started to see the following logs:
```
failed to send traces to intake at http://localhost:8126/v0.4/traces: HTTP error status 400, reason Bad Request, message b'msgp: attempted to decode type "float64" with method for "int" at 0/0/Start\n'
```

The problem comes from a misplaced int cast when setting the `start_ns` of the lambda-url inferred span.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/asm.py
+++ b/datadog_lambda/asm.py
@@ -1,0 +1,136 @@
+import logging
+from typing import Any, Dict, List, Optional
+
+from ddtrace.contrib.internal.trace_utils import _get_request_header_client_ip
+from ddtrace.internal import core
+from ddtrace.trace import Span
+
+from datadog_lambda.trigger import (
+    EventSubtypes,
+    EventTypes,
+    _EventSource,
+    _http_event_types,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _to_single_value_headers(request_headers: Dict[str, List[str]]) -> Dict[str, str]:
+    """
+    Convert multi-value headers to single-value headers.
+    If a header has multiple values, the first value is used.
+    """
+    single_value_headers = {}
+    for key, values in request_headers.items():
+        if len(values) >= 1:
+            single_value_headers[key] = values[0]
+    return single_value_headers
+
+
+def asm_start_request(
+    span: Span,
+    event: Dict[str, Any],
+    event_source: _EventSource,
+    trigger_tags: Dict[str, str],
+):
+    request_headers: Dict[str, str] = {}
+    peer_ip: Optional[str] = None
+    request_path_parameters: Optional[Dict[str, Any]] = None
+
+    if event_source.event_type == EventTypes.ALB:
+        headers = event.get("headers")
+        multi_value_request_headers = event.get("multiValueHeaders")
+        if multi_value_request_headers:
+            request_headers = _to_single_value_headers(multi_value_request_headers)
+        else:
+            request_headers = headers or {}
+
+        raw_uri = event.get("path")
+        parsed_query = event.get("multiValueQueryStringParameters") or event.get(
+            "queryStringParameters"
+        )
+
+    elif event_source.event_type == EventTypes.LAMBDA_FUNCTION_URL:
+        request_headers = event.get("headers", {})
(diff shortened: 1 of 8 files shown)


datadog_lambda/wrapper.py
```diff
@@ -285,6 +288,15 @@ def _after(self, event, context):
 
                 if status_code:
                     self.span.set_tag("http.status_code", status_code)
+
+                if config.appsec_enabled:
+                    asm_start_response(
+                        self.span,
+                        status_code,
+                        self.event_source,
+                        response=self.response,
```
joeyzhao2018, line 297: the `self.response` might not be a Dictionary, right?

florentinl: Yes you are right, it happens in two cases:
- malformed responses
- For API Gateway HTTP API events with payload version 2 or lambda function url events, the function may return just the body of the response

I pushed a fix, thank you.

datadog_lambda/asm.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+
+def asm_start_response(
+    span: Span,
+    status_code: str,
+    event_source: _EventSource,
+    response: Union[Dict[str, Any], str, None],
+):
+    if event_source.event_type not in _http_event_types:
```
joeyzhao2018, line 143: nitpick: personally I prefer adding this to `asm_start_request` too. Though it's currently doing the same checks but using the _http_event_types in both methods make it easier for the eyes IMHO.

joeyzhao2018 (approved):
LTGM

datadog_lambda/asm.py
```diff
@@ -0,0 +1,169 @@
...
+    body = event.get("body")
+    is_base64_encoded = event.get("isBase64Encoded", False)
+
+    request_ip = _get_request_header_client_ip(request_headers, peer_ip, True)
+    if request_ip is not None:
+        span.set_tag_str("http.client_ip", request_ip)
+        span.set_tag_str("network.client.ip", request_ip)
+
+    core.dispatch(
+        "aws_lambda.start_request",
```
joeyzhao2018, line 121: idea: maybe add a comment here to explain [the appsec listeners are registered in ddtrace](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/appsec/_handlers.py#L387)? so that people can connect the dots in the future? but let me know if you feel it's a bit too much

datadog_lambda/asm.py
florentinl: It makes sense, I added it.

datadog_lambda/asm.py
florentinl: Agreed, its better to have a comment to explain this. Done.

PR #622, opened 2025-06-20
feat: lambda support for DSM

michael-zhao459:
### What does this PR do?

This PR adds lambda support for Data Streams Monitoring (DSM) and reworks the original implementation. 

1) DSM context is passed through the trace propagation headers, code is refactored to use existing extraction logic (deleted dsm.py, reinventing the wheel here). 

2) If DSM is enabled, add custom DSM logic to the extracted context afterwards

### Motivation
Remove redundant code. DSM customers wanted to have Lambda support, currently context is not propagated correctly with lambdas. 

### Testing Guidelines

| Test case | Expected outcome | SQS  | SNS | SNS \-\> SQS (arn from SQS is used) | Kinesis |
| :---- | :---- | :---- | :---- | :---- | :---- |
| Datadog Context propagated properly through stringValue | Data streams context propagated & data streams checkpoint set |  ✅ | ✅ | ✅ | DNE |
| Datadog Context propagated properly through binaryValue | Data streams context propagated & data streams checkpoint set | ✅ | ✅ | ✅ | ✅ |
| No \_datadog message attribute  | Checkpoint set, no context propagation | ✅ | ✅ | ✅ | ✅ |
| Empty datadog message attribute | Checkpoint set, no context propagation | ✅ | ✅ | ✅ | ✅ |
| No data streams context in \_datadog message attribute | Checkpoint set, no context propagation | ✅ | ✅ | ✅ | ✅ |
| Invalid datadog message attribute | Checkpoint set, no context propagation, debug logger called | ✅ | ✅ | ✅ | ✅ |
| source\_arn is not found | No checkpoint set | ✅ | ✅ | ✅ | ✅ |
| Data streams disabled | No checkpoint set | ✅ | ✅ | ✅ | ✅ |

The tests go through all of the SQS case first, then all of the SNS case, then all of the SNS->SQS case, then all of the Kinesis case

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tracing.py
```diff
@@ -272,8 +307,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                         logger.debug(
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                if is_sqs:
+                    return _extract_context(dd_data, "sqs", arn)
+                return _extract_context(dd_data, "sns", sns_record.get("TopicArn", ""))
```
purple4reina: At this point, I know it's pretty impossible for `sns_record` not to be defined. But there is still the chance of it, especially in the future if someone makes changes. Here's what I'd recommend to protect against that.

Move the sns arn extraction to inside the if block where `sns_record` is defined. Then here, all you'll need to do is

```python
return _extract_context(dd_data, "sqs" if is_sqs else "sns", arn)
```

datadog_lambda/tracing.py
```diff
@@ -215,6 +245,8 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     Falls back to lambda context if no trace data is found in the SQS message attributes.
     """
+    is_sqs = False
+    arn = None
```
purple4reina: You technically don't need to define `arn` here since it's also defined below.

datadog_lambda/tracing.py
```diff
@@ -67,6 +66,37 @@
...
+    from ddtrace.data_streams import set_consume_checkpoint
+
+    """
+    Extracts the context from a JSON carrier and optionally sets a dsm consume checkpoint
+    if the context is complete and data streams are enabled.
+    """
+    context = propagator.extract(context_json)
+
+    if not _is_context_complete(context):
+        return context
```
purple4reina: Question: Is it guaranteed that if the dsm context is in the event, that the trace propagation headers will also be present?

datadog_lambda/tracing.py
```diff
@@ -67,6 +66,37 @@
...
+
+    if not config.data_streams_enabled:
+        return context
```
purple4reina: Since `_is_context_complete` takes a bit more work than checking `config.data_streams_enabled`, I'd reverse the order of these two if statements.

datadog_lambda/tracing.py
```diff
@@ -67,6 +66,37 @@
...
+    try:
+        carrier_get = _create_carrier_get(context_json)
```
purple4reina: Alternatively:

```python
carrier_get = lambda k: context_json.get(k)
```

datadog_lambda/tracing.py
michael-zhao459: Really good question. @piochelepiotr, is it fair to say that if the trace propagation headers are there dsm context will be there? I will see if I can figure it out myself

datadog_lambda/tracing.py
```diff
@@ -67,6 +66,37 @@
 LOWER_64_BITS = "LOWER_64_BITS"
 
 
+def _extract_context(context_json, event_type, arn):
```
purple4reina: I'm thinking we should update the name of this method to be a bit more specific. Now that I see it implemented, let's find a better name. I'm thinking something like `_extract_context_with_datastreams` or `_extract_context_and_dsm`.

datadog_lambda/tracing.py
```diff
@@ -226,6 +257,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     try:
         first_record = event.get("Records")[0]
+        arn = first_record.get("eventSourceARN", "")
+        if arn:
+            is_sqs = True
 
         # logic to deal with SNS => SQS event
         if "body" in first_record:
```
purple4reina, line 255: You do not currently have any testing for this block of code. Make sure to test the SNS => SQS case.

tests/test_dsm.py
```diff
@@ -1,112 +1,266 @@
...
+        self.addCleanup(logger_patcher.stop)
 
-        patcher = patch(
-            "ddtrace.internal.datastreams.botocore.calculate_sqs_payload_size"
-        )
-        self.mock_calculate_sqs_payload_size = patcher.start()
-        self.mock_calculate_sqs_payload_size.return_value = 100
-        self.addCleanup(patcher.stop)
+    def test_extract_context_data_streams_disabled(self):
+        with patch.dict(os.environ, {"DD_DATA_STREAMS_ENABLED": "false"}):
```
purple4reina: Alternatively so you don't need the extra level of indenting

```suggestion
    @patch("datadog_lambda.config.Config.data_streams_enabled", False)
    def test_extract_context_data_streams_disabled(self):
```

tests/test_dsm.py
```diff
@@ -1,112 +1,266 @@
...
+    _extract_context_with_data_streams,
+    _create_carrier_get,
+    extract_context_from_sqs_or_sns_event_or_context,
+    extract_context_from_kinesis_event,
+)
+from tests.utils import get_mock_context
 
-class TestDsmSQSContext(unittest.TestCase):
+
+class TestExtractContext(unittest.TestCase):
```
purple4reina: _Technically_ I think we'd wanna keep all these tests in `tests/test_tracing.py`. Usually we want the name of the test file to match the name of the file being tested.

datadog_lambda/tracing.py
michael-zhao459: Actually imo no matter what I think it makes more sense to explicitly check if DSM context is inside the context_json, I'll change this to be an explicit check for that instead

michael-zhao459: Piotr if you could help me check if https://github.com/DataDog/datadog-lambda-python/pull/622/commits/4d2fe091d351c498fa9a28e21795cbf40b1b8e8d this commit makes sense or not, thank you

tests/test_dsm.py
michael-zhao459: I agree that it's kinda breaking convention but I just think it'll be easier to maintain and makes more sense for the DSM specific tests to be in its own file

purple4reina: The benefit of having a 1:1 file to test is that when I make a change to a file, I know exactly which tests I will need to run while I'm doing my developement. I totally agree that the tracing file and it's tests are way too long. But let's think about splitting them up in a way that maintains the 1:1.

michael-zhao459: Alright makes 100% sense will do

datadog_lambda/tracing.py
michael-zhao459: @purple4reina Could you look over this change please as well to? https://github.com/DataDog/datadog-lambda-python/commit/4d2fe091d351c498fa9a28e21795cbf40b1b8e8d

datadog_lambda/tracing.py
```diff
@@ -67,6 +66,37 @@
 LOWER_64_BITS = "LOWER_64_BITS"
 
 
+def _extract_context_with_data_streams(context_json, event_type, arn):
```
piochelepiotr: setting the consume checkpoint in `extract_context_with_data_streams` function works, but this function feels like a read only function to me. So it's weird that it's actually setting a checkpoint.
Also, isn't this function used for more than sqs, sns & kinesis? We don't want to set checkpoints if the event type is a manual trigger for example. So I would call `set_consume_checkpoint` in the caller of `extract_context_with_data_streams`

datadog_lambda/wrapper.py
```diff
@@ -216,7 +215,13 @@ def _before(self, event, context):
 
             self.trigger_tags = extract_trigger_tags(event, context)
             # Extract Datadog trace context and source from incoming requests
-            dd_context, trace_context_source, event_source = extract_dd_trace_context(
+            (
+                dd_context,
+                trace_context_source,
+                event_source,
+                dd_json_data,
+                arn,
```
piochelepiotr: For the arn, I would prefer building it again in this function, instead of returning it. Let's only return the dd_json_data.

datadog_lambda/tracing.py
```diff
@@ -241,6 +245,8 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
         msg_attributes = first_record.get("messageAttributes")
         if msg_attributes is None:
             sns_record = first_record.get("Sns") or {}
+            if not is_sqs:
```
piochelepiotr: let's put this arn logic in the wrapper function

datadog_lambda/tracing.py
```diff
@@ -267,13 +274,16 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                            None,
+                            None,
+                        )
                     except Exception:
                         logger.debug(
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                return propagator.extract(dd_data), dd_data, arn
```
piochelepiotr: instead of returning the whole dd_data, which duplicates data from `propagator.extract(dd_data)`, let's return:
`{dsmCtxKey: dd_data[dsmCtxKey]} if dsmCtxKey in dd_data else {}`, what do you think? I believe the context key is available in the public library from dd-trace-py?

That way, you can call it `data_streams_ctx` in the upstream functions.

michael-zhao459: I do not really want to add anymore DSM constants into their file so I would prefer what the implementation looks like now

datadog_lambda/tracing.py
```diff
@@ -373,11 +378,11 @@ def extract_context_from_kinesis_event(event, lambda_context):
             data_obj = json.loads(data_str)
             dd_ctx = data_obj.get("_datadog")
             if dd_ctx:
-                return propagator.extract(dd_ctx)
+                return propagator.extract(dd_ctx), dd_ctx
```
piochelepiotr: this is not dd_ctx, it's data_streams_ctx

datadog_lambda/tracing.py
```diff
@@ -585,6 +590,8 @@ def extract_dd_trace_context(
     global dd_trace_context
     trace_context_source = None
     event_source = parse_event_source(event)
+    context = None
+    dd_json_data = None
```
piochelepiotr: same here, why is the name different? I think it should also be data_streams_ctx

datadog_lambda/tracing.py
```diff
@@ -585,6 +590,8 @@ def extract_dd_trace_context(
     global dd_trace_context
     trace_context_source = None
     event_source = parse_event_source(event)
+    context = None
```
piochelepiotr: why is context defined here now and not before?

datadog_lambda/tracing.py
michael-zhao459: At this point I think it still has all the datadog context, including the apm trace context, I just worry its a little misleading to call it data_streams_ctx. Please lemme know if you think I should still rename it, I just think we call it dd_ctx and then later on inside wrapper.py we extract out the DSM parts of it

datadog_lambda/tracing.py
michael-zhao459: I'll change this based on your feedback on the comment above

datadog_lambda/tracing.py
piochelepiotr: Oh, so let's not have the rest of the context in dd_ctx here.
Something like:
`data_streams_ctx = {dsmKey: dd_ctx[dsmKey]} if dsmKey in dd_ctx else {}`

and return that. That way, the function doesn't return data it doesn't need to return.

datadog_lambda/tracing.py
```diff
@@ -373,11 +389,20 @@ def extract_context_from_kinesis_event(event, lambda_context):
             data_obj = json.loads(data_str)
             dd_ctx = data_obj.get("_datadog")
             if dd_ctx:
-                return propagator.extract(dd_ctx)
+                if config.data_streams_enabled:
+                    from ddtrace.data_streams import PROPAGATION_KEY_BASE_64
+
+                    data_streams_ctx = {
```
michael-zhao459: I know the else is redundant but datadog gets mad if i just do the if too many indents

datadog_lambda/tracing.py
```diff
@@ -291,15 +303,18 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                                sampling_priority=float(x_ray_context["sampled"]),
+                            ),
+                            None,
                         )
-        return extract_context_from_lambda_context(lambda_context)
+        return extract_context_from_lambda_context(lambda_context), {}.get
     except Exception as e:
         logger.debug("The trace extractor returned with error %s", e)
-        return extract_context_from_lambda_context(lambda_context)
+        return extract_context_from_lambda_context(lambda_context), None
```
piochelepiotr: we should not return None here

piochelepiotr (approved):

datadog_lambda/tracing.py
```diff
@@ -265,15 +265,27 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                            extract_context_from_step_functions(dd_data, None),
+                            None,
+                        )
                     except Exception:
                         logger.debug(
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                return propagator.extract(dd_data), data_streams_ctx.get
```
joeyzhao2018, line 288: what exactly is the `.get` purpose here?

datadog_lambda/tracing.py
```diff
@@ -265,15 +265,27 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
             if dd_json_data:
                 dd_data = json.loads(dd_json_data)
 
+                data_streams_ctx = {}
+                if config.data_streams_enabled:
+                    from ddtrace.data_streams import PROPAGATION_KEY_BASE_64
```
joeyzhao2018: My main concerns are
1. Creating dictionary objects and bound methods for every invocation is inefficient.
2. It is very hard to follow the logic and hard to maintain and may introduce unexpected behaviors that are hard to debug in the future. 

May I suggest the following alternative implementation? Let me know what do you think.
```python

def _create_dsm_carrier_func(dd_data):
      """Create a carrier function for DSM context extraction."""
      def carrier_get(key):
          return dd_data.get(key) if dd_data else None
      return carrier_get

# then in In the extraction functions:
if config.data_streams_enabled:
    dsm_carrier = _create_dsm_carrier_func(dd_data)  # Pass the original dd_data
else:
    dsm_carrier = None
```

michael-zhao459: I agree with the justifications you made for this change will change the code now!

datadog_lambda/tracing.py
```diff
@@ -291,15 +293,18 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                        return (
+                            Context(
+                                trace_id=int(trace_id_parts[2][8:], 16),
+                                span_id=int(x_ray_context["parent_id"], 16),
+                                sampling_priority=float(x_ray_context["sampled"]),
+                            ),
+                            None,
                         )
-        return extract_context_from_lambda_context(lambda_context)
+        return extract_context_from_lambda_context(lambda_context), {}
```
joeyzhao2018: I think here it can just be `None`

datadog_lambda/tracing.py
joeyzhao2018: I can see that we have some back-and-forth on the decisions on what to return. This kind of confusion roots in the current code structure, which our team should be responsible to refactor later. So I think return dd_ctx this level here is fine for now. The future refactor task is to have the dd_ctx return first and then get related contexts by each specific methods. But that's out of the scope for this PR. Let me know what you all think.

datadog_lambda/tracing.py
```diff
@@ -291,15 +293,18 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
     except Exception as e:
         logger.debug("The trace extractor returned with error %s", e)
-        return extract_context_from_lambda_context(lambda_context)
+        return extract_context_from_lambda_context(lambda_context), {}
```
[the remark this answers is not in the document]
michael-zhao459: The idea here is that we want to set a DSM checkpoint even if we fail to extract the DSM context (produce call), we can still track the consume call, even if the produce call wasn't tracked.
So we could see:
{ queue } --> { service }
Instead of seeing nothing

datadog_lambda/tracing.py
joeyzhao2018: I just saw some conflicting comments from above. I think I'll make this non-blocking and approve the PR.

joeyzhao2018 (approved):
LGTM

datadog_lambda/tracing.py
michael-zhao459: The idea here is that we want to set a DSM checkpoint even if we fail to extract the DSM context (produce call), we can still track the consume call, even if the produce call wasn't tracked.
So we could see:
{ queue } --> { service }
Instead of seeing nothing

datadog_lambda/wrapper.py
```diff
@@ -357,4 +374,30 @@ def format_err_with_traceback(e):
...
+    return arn
+
+
+def _create_dsm_carrier_func(dd_data):
+    def carrier_get(key):
+        # {}.get(key) returns None, DSM checkpoint can be set even if DSM context extraction fails
+        # None.get(key) errors, preventing checkpoint for unsupported dsm context propagation method
+        return dd_data.get(key)
+
+    return carrier_get
```
purple4reina: All of this logic needs to be in the `tracing.py` file.

datadog_lambda/tracing.py
```diff
@@ -220,7 +220,11 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
     try:
         context = _extract_context_from_eventbridge_sqs_event(event)
         if _is_context_complete(context):
-            return context
+            return (
+                context,
+                # Unsupported DSM context propagation method
+                None,
+            )
```
purple4reina: What happened to the code we worked on together? How we created a new method to call instead of calling `propagator.extract`? That is the proper way to do this. We should not be changing the signature of all of these methods.

datadog_lambda/tracing.py
```diff
@@ -296,9 +331,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                             span_id=int(x_ray_context["parent_id"], 16),
                             sampling_priority=float(x_ray_context["sampled"]),
                         )
+        # Still want to set a DSM checkpoint even if DSM context not propagated
```
michael-zhao459, line 327: We want to set a checkpoint even when context is not propagated because we can still track the consume call, even if the produce call wasn't tracked.
So you would see:
{ queue } --> { service }
Instead of seeing nothing. It's still useful

datadog_lambda/tracing.py
```diff
@@ -67,6 +67,31 @@
 LOWER_64_BITS = "LOWER_64_BITS"
 
 
+def _dsm_set_checkpoint(context_json, event_type, arn):
+    from ddtrace.data_streams import set_consume_checkpoint
```
purple4reina: This should only be imported if it is going to be used. So, somewhere after the returns.

datadog_lambda/tracing.py
purple4reina: Can we use this definition instead? It'll be cleaner.

michael-zhao459: datadog_lambda/tracing.py:80:9: E731 do not assign a lambda expression, use a def
The way linting is setup currently lambda functions are no gos, can't pass the automated checks

purple4reina: You can override it. You'll add something like

```python
carrier_get = lambda k: context_json.get(k)  # noqa: E731
```

datadog_lambda/tracing.py
```diff
@@ -65,6 +65,28 @@
...
+
+def _dsm_set_checkpoint(context_json, event_type, arn):
+    if not config.data_streams_enabled:
+        return
+
+    if not isinstance(context_json, dict):
+        return
+
+    if context_json and PROPAGATION_KEY_BASE_64 not in context_json:
+        return
```
purple4reina, line 75: Is this really what you want? This will prevent a checkpoint from being set when there was context found, but your propagation key is not in it.

datadog_lambda/tracing.py
```diff
@@ -65,6 +65,28 @@
 DD_TRACE_JAVA_TRACE_ID_PADDING = "00000000"
 HIGHER_64_BITS = "HIGHER_64_BITS"
 LOWER_64_BITS = "LOWER_64_BITS"
+PROPAGATION_KEY_BASE_64 = "dd-pathway-ctx-base64"
```
purple4reina:
```suggestion
DSM_PROPAGATION_KEY_BASE_64 = "dd-pathway-ctx-base64"
```

datadog_lambda/tracing.py
```diff
@@ -202,7 +224,9 @@ def create_sns_event(message):
     }
 
 
-def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
+def extract_context_from_sqs_or_sns_event_or_context_and_set_dsm_ckpt_if_enabled(
```
purple4reina: I'm fine to just leave this method name what it was. I think the comment here suffices.

datadog_lambda/tracing.py
```diff
@@ -296,9 +328,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                             span_id=int(x_ray_context["parent_id"], 16),
                             sampling_priority=float(x_ray_context["sampled"]),
                         )
+        # Still want to set a DSM checkpoint even if DSM context not propagated
+        _dsm_set_checkpoint({}, "sqs" if is_sqs else "sns", arn)
```
purple4reina: Avoid the needless allocation.

```suggestion
        _dsm_set_checkpoint(None, "sqs" if is_sqs else "sns", arn)
```

datadog_lambda/tracing.py
```diff
@@ -296,9 +328,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
         return extract_context_from_lambda_context(lambda_context)
     except Exception as e:
         logger.debug("The trace extractor returned with error %s", e)
+        # Still want to set a DSM checkpoint even if DSM context not propagated
+        _dsm_set_checkpoint({}, "sqs" if is_sqs else "sns", arn)
```
purple4reina:
```suggestion
        _dsm_set_checkpoint(None, "sqs" if is_sqs else "sns", arn)
```

datadog_lambda/tracing.py
```diff
@@ -226,6 +252,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
 
     try:
         first_record = event.get("Records")[0]
```
purple4reina, line 251: You should define `arn = None` or `arn = ""` above as you did with `is_sqs`.

In the case where this line fails, `arn` will not yet be defined leading to an exception being raised during the exception handling.

datadog_lambda/tracing.py
```diff
@@ -354,12 +390,16 @@ def extract_context_from_eventbridge_event(event, lambda_context):
         return extract_context_from_lambda_context(lambda_context)
 
 
-def extract_context_from_kinesis_event(event, lambda_context):
+def extract_context_from_kinesis_event_and_set_dsm_checkpoint_if_enabled(
```
purple4reina: Likewise, I think we should revert this back to the original method name.

datadog_lambda/tracing.py
```diff
@@ -373,10 +413,13 @@ def extract_context_from_kinesis_event(event, lambda_context):
...
             if dd_ctx:
-                return propagator.extract(dd_ctx)
+                context = propagator.extract(dd_ctx)
+                _dsm_set_checkpoint(dd_ctx, "kinesis", arn)
+                return context
     except Exception as e:
         logger.debug("The trace extractor returned with error %s", e)
-
+    # Still want to set a DSM checkpoint even if DSM context not propagated
+    _dsm_set_checkpoint({}, "kinesis", arn)
```
purple4reina:
```suggestion
    _dsm_set_checkpoint(None, "kinesis", arn)
```

datadog_lambda/tracing.py
```diff
@@ -354,12 +390,16 @@ def extract_context_from_eventbridge_event(event, lambda_context):
...
+def extract_context_from_kinesis_event_and_set_dsm_checkpoint_if_enabled(
+    event, lambda_context
+):
     """
     Extract datadog trace context from a Kinesis Stream's base64 encoded data string
+    Set a DSM checkpoint if DSM is enabled and the method for context propagation is supported.
     """
     try:
         record = get_first_record(event)
+        arn = record.get("eventSourceARN", "")
```
purple4reina: Likewise, if an exception is raised on the line above, the `arn` will not be defined below.

purple4reina (changes requested):
All of the comments I've left below indicate that there is missing test coverage. Make sure to include tests for all corner cases.

datadog_lambda/tracing.py
michael-zhao459: def set_consume_checkpoint(typ, source, carrier_get):
      processor.decode(carrier_get(PROPAGATION_KEY_BASE_64))
      processor.set_checkpoint()


This is the psuedocode for the checkpoint code. The idea of sending a {} is that {}.get(KEY) will return None, which in the decode is still allowed and will set a checkpoint. If I send None None.get(KEY) will throw an exception and no checkpoint will be set

purple4reina: Then you should change your carrier get method to handle the None. That will avoid the extra allocation.

datadog_lambda/tracing.py
```diff
@@ -65,6 +65,28 @@
...
+    if not isinstance(context_json, dict):
+        return
+
+    if context_json and PROPAGATION_KEY_BASE_64 not in context_json:
+        return
+
+    try:
+        from ddtrace.data_streams import set_consume_checkpoint
+
+        carrier_get = lambda k: context_json.get(k)  # noqa: E731
```
purple4reina:
```suggestion
        carrier_get = lambda k: context_json and context_json.get(k)  # noqa: E731
```

datadog_lambda/tracing.py
michael-zhao459: This is what is wanted. I want either a {} or context_json with the key, data_streams_ctx = {dsmKey: dd_ctx[dsmKey]} if dsmKey in dd_ctx else {} was actually suggested by Piotr in a earlier implementation. I do want to check that the dsmKey is there

purple4reina: Gotcha. But in the same way that a checkpoint is helpful when the context is empty, wouldn't it be helpful to set a checkpoint when it isn't empty but the dsm key isn't present? Might be a nice to have.

datadog_lambda/tracing.py
michael-zhao459: Got it!

datadog_lambda/tracing.py
michael-zhao459: lemme check with Piotr

datadog_lambda/tracing.py
```diff
@@ -373,10 +421,15 @@ def extract_context_from_kinesis_event(event, lambda_context):
             data_obj = json.loads(data_str)
             dd_ctx = data_obj.get("_datadog")
```
piochelepiotr, line 409: when would that be set for Kinesis?

datadog_lambda/tracing.py
```diff
@@ -67,6 +67,21 @@
...
 
 
+def _dsm_set_checkpoint(context_json, event_type, arn):
+    if not config.data_streams_enabled:
+        return
+
+    try:
+        from ddtrace.data_streams import set_consume_checkpoint
+
+        carrier_get = lambda k: context_json and context_json.get(k)  # noqa: E731
```
michael-zhao459, line 80: Ok for context_json to not have DSM_KEY. Additional note is not checking if context_json is a dict before processing, datadog context should be injected as a dict and already placed in try block for assurance

datadog_lambda/tracing.py
```diff
@@ -241,6 +260,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
         msg_attributes = first_record.get("messageAttributes")
         if msg_attributes is None:
             sns_record = first_record.get("Sns") or {}
+            # SNS->SQS event would extract SNS arn without this check
+            if event_source.equals(EventTypes.SNS):
```
michael-zhao459, line 264: Without this check, in the SNS -> SQS case the arn would get changed to the SNS topic arn, I want to maintain the SQS arn extracted beforehand

datadog_lambda/tracing.py
```diff
@@ -272,8 +294,16 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                         logger.debug(
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                context = propagator.extract(dd_data)
+                # Do not want to set checkpoint with "" arn
+                if arn:
```
michael-zhao459: A "" arn would end up with a queue with no name, causing many collisions and overall bad behavior. Only set a checkpoint if arn has a non empty string

datadog_lambda/tracing.py
```diff
@@ -214,7 +229,10 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
     Lambda Context.
 
     Falls back to lambda context if no trace data is found in the SQS message attributes.
+    Set a DSM checkpoint if DSM is enabled and the method for context propagation is supported.
     """
+    arn = ""
+    event_source = parse_event_source(event)
```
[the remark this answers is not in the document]
michael-zhao459: This line can also be written as arn and set_consume_checkpoint(), please let me know if that is preferred Rey

datadog_lambda/tracing.py
michael-zhao459: This line can also be written as arn and set_consume_checkpoint(), let me know if this is preferred @purple4reina

datadog_lambda/tracing.py
michael-zhao459: https://github.com/DataDog/dd-trace-py/blob/926d8383af8e71c6a83494adf85918a4ad7cf920/ddtrace/internal/datastreams/botocore.py#L230 Where we inject DSM context for Kinesis produce call

datadog_lambda/tracing.py
```diff
@@ -272,8 +295,16 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
                         logger.debug(
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                context = propagator.extract(dd_data)
+                # Do not want to set checkpoint with "" arn
+                if source_arn:
```
purple4reina: I would push this source arn checking to the `_dsm_set_checkpoint` method itself. That way you only need to do the check in one location instead of five.

datadog_lambda/tracing.py
```diff
@@ -272,8 +295,16 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
+                    _dsm_set_checkpoint(
+                        dd_data,
+                        # In this function only recieves SQS and SNS events, if not SQS must be SNS
```
purple4reina: I think we can remove this comment. It's obvious enough what's going on without it.

datadog_lambda/tracing.py
```diff
@@ -241,6 +261,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
         msg_attributes = first_record.get("messageAttributes")
         if msg_attributes is None:
             sns_record = first_record.get("Sns") or {}
+            # SNS->SQS event would extract SNS arn without this check
```
purple4reina, line 268: Just want to double check that this is what you want.

If we have something like Lambda -> SNS -> SQS -> Lambda, do we set the checkpoint using the SNS arn or the SQS arn?

I would think we would set it with the SNS arn since that's the arn we would be setting it with on the produce end. But I don't know DSM well enough to know. @piochelepiotr do you know?

datadog_lambda/tracing.py
```diff
@@ -214,7 +231,9 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
     Lambda Context.
 
     Falls back to lambda context if no trace data is found in the SQS message attributes.
+    Set a DSM checkpoint if DSM is enabled and the method for context propagation is supported.
     """
+    source_arn = ""
```
purple4reina, line 239: I think it would be worth it to add another variable declaration here for 

```python
event_type = "sqs" if event_source.equals(EventTypes.SQS) else "sns"
```

That way we only need to manage this logic in one location instead of three.

datadog_lambda/tracing.py
piochelepiotr: we want to use the SQS arn in this case. Otherwise, we want be able to know that data went through sqs.

Data streams will handle that correctly in the backend by using the context that was propagated (so it knows who the parent is)

purple4reina: Gotcha, thanks for the clarification!!

datadog_lambda/tracing.py
```diff
@@ -272,8 +299,13 @@ def extract_context_from_sqs_or_sns_event_or_context(event, lambda_context):
...
                             "Failed to extract Step Functions context from SQS/SNS event."
                         )
-
-                return propagator.extract(dd_data)
+                context = propagator.extract(dd_data)
+                _dsm_set_checkpoint(
+                    dd_data,
+                    event_type,
+                    source_arn,
+                )
```
purple4reina: nit: this can all be on a single line now.

michael-zhao459: When I do black datadog_lambda/tracing.py this is how it wants me to format it, is there anyway I should be overriding this if you want me to?

purple4reina: Just put them all on a single row. Black won't change them back.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 python = ">=3.8.0,<4"
 datadog = ">=0.51.0,<1.0.0"
 wrapt = "^1.11.2"
-ddtrace = ">=2.20.0,<4"
+ddtrace = ">=3.10.0"
```
piochelepiotr: this changes the major version. Is that what we want to do? Also, should we keep `<4`?

tests/test_tracing.py
```diff
@@ -2438,3 +2443,905 @@ def test_exception_outside_handler_tracing_disabled(
...
+        extract_context_from_sqs_or_sns_event_or_context(
+            event, self.lambda_context, parse_event_source(event)
+        )
+        # None indiciates no DSM context propagation
+        mock_dsm_set_checkpoint.assert_called_once_with(
+            None, "sqs", "arn:aws:sqs:us-east-1:123456789012:test-queue"
+        )
+
+    @patch("datadog_lambda.tracing._dsm_set_checkpoint")
+    def test_sqs_incorrect_datadog_message_attribute(self, mock_dsm_set_checkpoint):
```
piochelepiotr: nit: `incorrect` -> `invalid`

tests/test_tracing.py
```diff
@@ -2438,3 +2443,905 @@ def test_exception_outside_handler_tracing_disabled(
...
+        )
+
+        # DSM context is not propagated
+        context_json = mock_dsm_set_checkpoint.call_args[0][0]
+        carrier_get = lambda k: context_json and context_json.get(k)  # noqa: E731
+        assert carrier_get("dd-pathway-ctx-base64") is None
+
+    @patch("datadog_lambda.tracing._dsm_set_checkpoint")
+    @patch("datadog_lambda.tracing.logger")
+    def test_sqs_invalid_datadog_message_attribute_raises_exception(
```
piochelepiotr: nit: remove `raises_exception` from the test name. We care about what the test case tests, not the logic under the hood. Maybe in the future, the code won't raise an exception, and it's still OK because the function accepts invalid datadog message attributes.

tests/test_tracing.py
```diff
@@ -2438,3 +2443,905 @@ def test_exception_outside_handler_tracing_disabled(
...
+                    "eventSource": "aws:sqs",
+                }
+            ]
+        }
+
+        extract_context_from_sqs_or_sns_event_or_context(
+            event, self.lambda_context, parse_event_source(event)
+        )
+
+        mock_dsm_set_checkpoint.assert_called_once_with(None, "sqs", "")
```
piochelepiotr: I don't understand, are we testing `mock_dsm_set_checkpoint`, or `self.mock_checkpoint`. We should test only one of them, and it should be consistent across all tests. If possible, it's better to test the lower level one (self.mock_checkpoint)

pyproject.toml
michael-zhao459: My mistake on the <4. The code will break without ddtrace version 3.10.0.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 python = ">=3.8.0,<4"
 datadog = ">=0.51.0,<1.0.0"
 wrapt = "^1.11.2"
-ddtrace = ">=2.20.0,<4"
+ddtrace = ">=3.10.0,<4"
```
purple4reina: We're currently working on a v3.10.2 release, once that's out, we should update this.

purple4reina (approved):

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 python = ">=3.8.0,<4"
 datadog = ">=0.51.0,<1.0.0"
 wrapt = "^1.11.2"
-ddtrace = ">=2.20.0,<4"
+ddtrace = ">=3.10.2,<4"
```
purple4reina, line 31: Should be good to release this now that v3.10.2 is up in pypi.

PR #630, opened 2025-07-10
feat(appsec): enable request blocking

florentinl:
### What does this PR do?

This PR enables Appsec to block request if they are deemed dangerous by the in-app WAF.

There are three cases:
- Blocking before executing the wrapped function, after calling `asm_start_request`
- Blocking during execution of the wrapped function, by raising a `BlockingException`
- Blocking after execution of the function,  after calling `asm_start_response`

### Motivation

Expand the capabilities of Appsec inside the datadog-lambda-python layer.

### Testing Guidelines

I added unit tests to both `test_asm.py` and `test_wrapper.py` to ensure requests are blocked only if necessary and at the right stage.

Integration tests will be run as part of the system-test, APPSEC_BLOCKING scenario.

### Bug Fix

The appsec waf expects the raw URI to contain the query parameters. This is an omission from the previous Appsec PRs. The fix is included in this one.

Some rules in the system-tests use the presence of query parameters in the URI to block requests. This is therefore required to pass the `APPSEC_BLOCKING` test suite of the system tests which this PR relates to.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/asm.py
+++ b/datadog_lambda/asm.py
@@ -1,9 +1,12 @@
-from copy import deepcopy
 import logging
+import urllib.parse
+from copy import deepcopy
 from typing import Any, Dict, List, Optional, Union
 
 from ddtrace.contrib.internal.trace_utils import _get_request_header_client_ip
 from ddtrace.internal import core
+from ddtrace.internal.utils import get_blocked
+from ddtrace.internal.utils import http as http_utils
 from ddtrace.trace import Span
 
 from datadog_lambda.trigger import (
@@ -50,6 +53,7 @@ def asm_set_context(event_source: _EventSource):
     This allows the AppSecSpanProcessor to know information about the event
     at the moment the span is created and skip it when not relevant.
     """
+
     if event_source.event_type not in _http_event_types:
         core.set_item("appsec_skip_next_lambda_event", True)
 
@@ -126,6 +130,14 @@ def asm_start_request(
         span.set_tag_str("http.client_ip", request_ip)
         span.set_tag_str("network.client.ip", request_ip)
 
+    # Encode the parsed query and append it to reconstruct the original raw URI expected by AppSec.
+    if parsed_query:
+        try:
+            encoded_query = urllib.parse.urlencode(parsed_query, doseq=True)
+            raw_uri += "?" + encoded_query  # type: ignore
+        except Exception:
+            pass
+
     core.dispatch(
         # The matching listener is registered in ddtrace.appsec._handlers
         "aws_lambda.start_request",
@@ -182,3 +194,36 @@ def asm_start_response(
             response_headers,
         ),
     )
+
+
+def get_asm_blocked_response(
+    event_source: _EventSource,
+) -> Optional[Dict[str, Any]]:
+    """Get the blocked response for the given event source."""
+    if event_source.event_type not in _http_event_types:
+        return None
+
+    blocked = get_blocked()
+    if not blocked:
+        return None
+
+    desired_type = blocked.get("type", "auto")
+    if desired_type == "none":
+        content_type = "text/plain; charset=utf-8"
+        content = ""
+    else:
+        content_type = blocked.get("content-type", "application/json")
+        content = http_utils._get_blocked_template(content_type)
+
+    response_headers = {
+        "content-type": content_type,
+    }
+    if "location" in blocked:
+        response_headers["location"] = blocked["location"]
+
+    return {
+        "statusCode": blocked.get("status_code", 403),
+        "headers": response_headers,
+        "body": content,
+        "isBase64Encoded": False,
+    }
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -9,7 +9,7 @@ import ujson as json
 from importlib import import_module
 from time import time_ns
 
-from datadog_lambda.asm import asm_set_context, asm_start_response, asm_start_request
+from ddtrace.internal._exceptions import BlockingException
 from datadog_lambda.extension import should_use_extension, flush_extension
 from datadog_lambda.cold_start import (
     set_cold_start,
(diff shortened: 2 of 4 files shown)


datadog_lambda/asm.py
```diff
@@ -126,6 +130,14 @@ def asm_start_request(
         span.set_tag_str("http.client_ip", request_ip)
         span.set_tag_str("network.client.ip", request_ip)
 
+    # Encode the parsed query and append it to reconstruct the original raw URI expected by AppSec.
+    if parsed_query:
+        try:
+            encoded_query = urllib.parse.urlencode(parsed_query, doseq=True)
+        except Exception:
+            pass
+        raw_uri += "?" + encoded_query  # type: ignore
```
purple4reina: If an exception is raised when url encoding the query, then the `encoded_query` will not be defined here.

purple4reina: Here's one way to better handle this.

```suggestion
        try:
            encoded_query = urllib.parse.urlencode(parsed_query, doseq=True)
            raw_uri += "?" + encoded_query  # type: ignore
        except Exception:
            pass
```

purple4reina (commented):
One small issue to fix.

datadog_lambda/asm.py
florentinl: You are right, I moved it inside as you suggested. Thank you !

datadog_lambda/asm.py
```diff
@@ -126,6 +130,14 @@ def asm_start_request(
         span.set_tag_str("http.client_ip", request_ip)
         span.set_tag_str("network.client.ip", request_ip)
 
+    # Encode the parsed query and append it to reconstruct the original raw URI expected by AppSec.
+    if parsed_query:
```
joeyzhao2018, line 134: Could you please explain a bit in the PR description how this part is related? Thank you

datadog_lambda/wrapper.py
```diff
@@ -155,12 +162,21 @@ def __init__(self, func):
         except Exception as e:
             logger.error(format_err_with_traceback(e))
 
+    def _get_blocking_response(self):
+        if not config.appsec_enabled:
```
joeyzhao2018: I think we might not need this method at all. In its 3 usages, 2 actually have the `config.appsec_enabled` right before calling it anyway. The only place left is in the `BlockingException` case, which implies that appsec already enabled? Or we just add the config.appsec_enabled check there? 
in short, I think we can just call `get_asm_blocked_response(self.event_source)` directly in the 3 usages so far to avoid some double `if` statements..

datadog_lambda/wrapper.py
```diff
@@ -9,7 +9,13 @@
...
 from time import time_ns
 
-from datadog_lambda.asm import asm_set_context, asm_start_response, asm_start_request
+from ddtrace.internal._exceptions import BlockingException
+from datadog_lambda.asm import (
+    asm_set_context,
+    asm_start_response,
+    asm_start_request,
+    get_asm_blocked_response,
+)
```
joeyzhao2018:
```suggestion
if config.appsec_enabled:
    from datadog_lambda.asm import (
        asm_set_context,
        asm_start_response,
        asm_start_request,
        get_asm_blocked_response,
    )
```
Add move it below `from datadog_lambda.config import config`

datadog_lambda/asm.py
```diff
@@ -182,3 +194,37 @@ def asm_start_response(
...
+    event_source: _EventSource,
+) -> Optional[Dict[str, Any]]:
+    """Get the blocked response for the given event source."""
+    if event_source.event_type not in _http_event_types:
+        return None
+
+    blocked = get_blocked()
+    if not blocked:
+        return None
+    set_blocked(blocked)
```
joeyzhao2018: i need some help understanding this part. Since we already got the blocked above, does the `set_blocked` have any effect here?

tests/test_asm.py
```diff
@@ -199,6 +208,30 @@
     ),
 ]
 
+ASM_BLOCKED_RESPONSE_TEST_CASES = [
```
joeyzhao2018, line 211: Claude gave me these following cases below. Could you add them here if they make sense? Thank you.
```
 # Default blocking configuration
  {
      "status_code": 403,
      "type": "auto",
      "content-type": "application/json"  # or "text/html" based on Accept header
  }

  # HTML blocking response
  {
      "status_code": 403,
      "type": "html",
      "content-type": "text/html"
  }

  # Redirect blocking
  {
      "status_code": 403,
      "type": "none",
      "location": "https://example.com/blocked"
  }
```

datadog_lambda/asm.py
florentinl: Done

> ### Bug Fix
> 
> The appsec waf expects the raw URI to contain the query parameters. This is an omission from the previous Appsec PRs. The fix is included in this one.
> 
> Some rules in the system-tests use the presence of query parameters in the URI to block requests. This is therefore required to pass the `APPSEC_BLOCKING` test suite of the system tests which this PR relates to.

tests/test_asm.py
florentinl: I cleaned up the test cases by adding yours and removing duplicates.

datadog_lambda/asm.py
florentinl: I thought that calling set_blocked added the expected `content-type` calculated from the accept headers gathered by Appsec. In fact it is already done by the WAF and unnecessary outside of certain frameworks.
I removed the set_blocked line as it is useless.

datadog_lambda/wrapper.py
florentinl: Done. This also implies some changes in the `test_wapper.py` where I have to reload the import to test appsec.

datadog_lambda/wrapper.py
florentinl: Yes it makes total sense to avoid checking it twice.
For the `BlockingException` case, it can only be raised when the ASM context is set therefore only when appsec is already enabled. No need to do the check in this case either.

PR #634, opened 2025-07-17
v7.113.0-dev

joeyzhao2018:
### What does this PR do?

Tag current code with a `devN` version indicating developing for next release.

### Motivation

This is needed for APM System testing. 

### Additional Notes

PEP 440 normalization: The standard requires that development releases use the .devN format (where N is a number, defaulting to 0 if not specified)

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


florentinl (approved):

florentinl:
Hey @duncanista and @joeyzhao2018,

The reason I need this is entirely system-test related, and I don't want to impose a burden on the release process of this repo for something that only appsec will ever use (I assume). Let me explain why I need this and we can probably find a better solution 😅 

`system-tests` work using manifests that declare the first version of a library that pass a compliance test. They look like this:  https://github.com/DataDog/system-tests/blob/main/manifests/python.yml

To know which tests to run or not (or more accurately to declare as XFailed or XPassed) it compares the version of the running library against this manifest. To do this it uses the self reported version of the package (`datadog_lambda.__version__` for us).

Without pre-release version markers, If the main branch passes a new test there is no way to currently mark the test as passing until the next release. 
For example if a release (let say v7.112.0), a new commit makes a new test pass, I can't mark it as passed for `v7.112.0` because running system-tests on the latest release would fail and I can't mark it as passed for `v7.113.0` because the test wouldn't be executed for the current main (also v7.113.0 might never exist because of a breaking change) and a regression could occur in a successive commit.

Having a `v7.113.0.devN` solves the issue by allowing to mark a test as passing for unreleased versions of a library.


If modifying the version manually inside the repo is too much of a burden, maybe we can try to automate it. `dd-trace-py` uses `setuptools_scm` there must be an equivalent for poetry. I can take a look if you want.

Another simpler approach would be to patch this directly in the artifact built for system-tests here https://github.com/DataDog/datadog-lambda-python/blob/main/.github/workflows/build_layer.yml. It makes running local builds against system-tests less straight-forward but I guess it is the path of least resistance.

florentinl:
To clarify this point
> my preliminary understanding is that those tests are using very limited resources and we are trying to make thing more directly accessible in the pipeline.

system-tests aren't really run with limited resource. My understanding from discussing with the system-tests team is that it is more of a problem of separation of concerns. 

They work under the principle that system-tests shouldn't have to worry about how to build a library or derive a version depending on the commit it comes from, these are the concerns of the library. system-tests only need to know where to fetch the library from and then use it as a Datadog user would.

PR #643, opened 2025-07-30
fix(dsm): set dsm checkpoint for all records in event

michael-zhao459:
### What does this PR do?

This PR 

Sets a DSM checkpoint for every single record in a event. Does this by extracting out a helper that gets datadog context per record, and if DSM is enabled continues to loop starting with the second record and setting DSM checkpoints (always using the first record for APM).

### Motivation

Screenshot 2025-07-30 at 2 19 19 PM

Please note the discrepancy between the msg/s from incoming produce and outgoing by downstream queue. When batch processing, if the consume service does not set a checkpoint for each message coming from upstream we lose track of them causing the noticeable drop in throughput. 

 
### Testing Guidelines

All of the tests in https://github.com/DataDog/datadog-lambda-python/pull/622 this table are maintained. Ensured no regressions by making sure that all test_tracing.py continued to pass.  Tested on sandbox AWS account for all queue types to see context propagation and ensure throughput matches. Added tests to show that DSM can now handle multiple records in a event. 

### Types of Changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -2743,6 +2743,73 @@ class TestExtractDDContextWithDSMLogic(unittest.TestCase):
         carrier_get_2 = args_2[2]
         self.assertEqual(carrier_get_2("dd-pathway-ctx-base64"), "record2")
 
+    def test_sqs_batch_processing_with_invalid_records(self):
+        dd_data_1 = {"dd-pathway-ctx-base64": "valid_record"}
+        dd_json_data_1 = json.dumps(dd_data_1)
+
+        dd_data_3 = {"dd-pathway-ctx-base64": "another_valid_record"}
+        dd_json_data_3 = json.dumps(dd_data_3)
+
+        event = {
+            "Records": [
+                {
+                    "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:test-queue",
+                    "messageAttributes": {
+                        "_datadog": {
+                            "dataType": "String",
+                            "stringValue": dd_json_data_1,
+                        }
+                    },
+                    "eventSource": "aws:sqs",
+                },
+                {
+                    "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:test-queue",
+                    "messageAttributes": {
+                        "_datadog": {
+                            "dataType": "Binary",
+                            # This will cause extraction to fail
+                            "binaryValue": "invalid-base64-data",
+                        }
+                    },
+                    "eventSource": "aws:sqs",
+                },
+                {
+                    "eventSourceARN": "arn:aws:sqs:us-east-1:123456789012:test-queue",
+                    "messageAttributes": {
+                        "_datadog": {
+                            "dataType": "String",
+                            "stringValue": dd_json_data_3,
+                        }
+                    },
+                    "eventSource": "aws:sqs",
+                },
+            ]
+        }
+
+        extract_context_from_sqs_or_sns_event_or_context(
+            event, self.lambda_context, parse_event_source(event)
+        )
+
+        self.assertEqual(self.mock_checkpoint.call_count, 3)
+
+        args_1, _ = self.mock_checkpoint.call_args_list[0]
+        self.assertEqual(args_1[0], "sqs")
+        self.assertEqual(args_1[1], "arn:aws:sqs:us-east-1:123456789012:test-queue")
+        carrier_get_1 = args_1[2]
+        self.assertEqual(carrier_get_1("dd-pathway-ctx-base64"), "valid_record")
+
+        args_2, _ = self.mock_checkpoint.call_args_list[1]
+        self.assertEqual(args_2[0], "sqs")
+        self.assertEqual(args_2[1], "arn:aws:sqs:us-east-1:123456789012:test-queue")
+        carrier_get_2 = args_2[2]
+        self.assertEqual(carrier_get_2("dd-pathway-ctx-base64"), None)
+
+        args_3, _ = self.mock_checkpoint.call_args_list[2]
+        self.assertEqual(args_3[0], "sqs")
+        self.assertEqual(args_3[1], "arn:aws:sqs:us-east-1:123456789012:test-queue")
+        carrier_get_3 = args_3[2]
+        self.assertEqual(carrier_get_3("dd-pathway-ctx-base64"), "another_valid_record")
+
     def test_sqs_source_arn_not_found(self):
         event = {
             "Records": [
@@ -3018,53 +3085,6 @@ class TestExtractDDContextWithDSMLogic(unittest.TestCase):
         _dsm_set_checkpoint(context_json, event_type, arn)
 
         self.mock_checkpoint.assert_not_called()
-
-    def test_sns_batch_processing(self):
-        dd_data_1 = {"dd-pathway-ctx-base64": "record1"}
-        dd_data_2 = {"dd-pathway-ctx-base64": "record2"}
-        dd_json_data_1 = json.dumps(dd_data_1)
-        dd_json_data_2 = json.dumps(dd_data_2)
-
-        event = {
-            "Records": [
-                {
-                    "Sns": {
-                        "TopicArn": "arn:aws:sns:us-east-1:123456789012:test-topic",
-                        "MessageAttributes": {
-                            "_datadog": {"Type": "String", "Value": dd_json_data_1}
-                        },
-                    },
-                    "eventSource": "aws:sns",
-                },
-                {
-                    "Sns": {
-                        "TopicArn": "arn:aws:sns:us-east-1:123456789012:test-topic",
-                        "MessageAttributes": {
-                            "_datadog": {"Type": "String", "Value": dd_json_data_2}
-                        },
-                    },
-                    "eventSource": "aws:sns",
-                },
-            ]
-        }
-
-        extract_context_from_sqs_or_sns_event_or_context(
-            event, self.lambda_context, parse_event_source(event)
-        )
-
-        self.assertEqual(self.mock_checkpoint.call_count, 2)
-
-        args_1, _ = self.mock_checkpoint.call_args_list[0]
-        self.assertEqual(args_1[0], "sns")
-        self.assertEqual(args_1[1], "arn:aws:sns:us-east-1:123456789012:test-topic")
-        carrier_get_1 = args_1[2]
-        self.assertEqual(carrier_get_1("dd-pathway-ctx-base64"), "record1")
-
-        args_2, _ = self.mock_checkpoint.call_args_list[1]
-        self.assertEqual(args_2[0], "sns")
-        self.assertEqual(args_2[1], "arn:aws:sns:us-east-1:123456789012:test-topic")
-        carrier_get_2 = args_2[2]
-        self.assertEqual(carrier_get_2("dd-pathway-ctx-base64"), "record2")
 
     # SNS -> SQS TESTS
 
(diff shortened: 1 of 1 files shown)


tests/test_dsm.py
```diff
@@ -0,0 +1,1048 @@
...
+
+    @patch("datadog_lambda.config.Config.data_streams_enabled", False)
+    def test_sqs_data_streams_disabled(self):
+        context_json = {"dd-pathway-ctx-base64": "12345"}
+
+        set_dsm_context(context_json, _EventSource(EventTypes.SQS))
+
+        self.mock_checkpoint.assert_not_called()
+
+    def test_sqs_multiple_records(self):
```
michael-zhao459, line 221: Extra test I added for all propagation types showing that I can now process each record

datadog_lambda/tracing.py
```diff
@@ -68,24 +68,6 @@
 LOWER_64_BITS = "LOWER_64_BITS"
 
 
-def _dsm_set_checkpoint(context_json, event_type, arn):
```
michael-zhao459, line 71: I think its the right thing to do to decouple the DSM logic from the trace extraction.

datadog_lambda/dsm.py
```diff
@@ -0,0 +1,60 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def set_dsm_context(event, event_source: _EventSource):
+    if (
+        not event_source.equals(EventTypes.KINESIS)
+        and not event_source.equals(EventTypes.SNS)
+        and not event_source.equals(EventTypes.SQS)
+    ):
```
piochelepiotr: do we expect this to happen? If not, let's add a debug log here.

datadog_lambda/tracing.py
```diff
@@ -250,59 +227,17 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
-                dd_json_data = dd_payload.get("stringValue") or dd_payload.get("Value")
-            else:
-                logger.debug(
-                    "Datadog Lambda Python only supports extracting trace"
-                    "context from String or Binary SQS/SNS message attributes"
-                )
-
-            if dd_json_data:
-                dd_data = json.loads(dd_json_data)
+        dd_data = extract_datadog_context_from_sqs_or_sns_event(first_record)
```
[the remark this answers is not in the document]
michael-zhao459: What I intended to convey with `extract_datadog_context_from_sqs_or_sns_event` is that it extracts the _datadog field. That way I can reuse this extract _datadog field for all the records in DSM, and extract _datadog field once here for APM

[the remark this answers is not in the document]
michael-zhao459: I should be naming functions not after internal functionality but based on what it does. Something like `extract_context_from_sqs_or_sns_record` would make more sense

datadog_lambda/dsm.py
joeyzhao2018: I agree the double de-serialization introduce some performance cost. But since we only need 1 extra deserialization for tracing purpose, and one JSON.loads(50 bytes) is at the level of nanosecond cost. So I feel comfortable not worrying about it at all. I'd say decoupling and better code structure is worth more. Also for lambdas, cold start costs are more critical than this. With that being said,...
1. If in the future we need to deserialize all the tracecontext for spanlinks for example, that would bring the cost to ms level for each invocation and we might want to refactor the code to do it only once then.
2. Might be too late to mention, but I start to think maybe datadog-lambda-js is a better starting place for refactoring like this because there we are using a bunch of extractors and could be easier to refactor.

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,102 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                            )
+                    if is_first_record:
+                        context = propagator.extract(dd_data)
+                    dsm_data = dd_data
             else:
-                logger.debug(
-                    "Datadog Lambda Python only supports extracting trace"
-                    "context from String or Binary SQS/SNS message attributes"
-                )
+                # Handle case where trace context is injected into attributes.AWSTraceHeader
```
michael-zhao459, line 317: Checked in dd-trace-py, DSM never injects context in here

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,102 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                    )
 
-                dd_json_data = dd_payload.get("binaryValue") or dd_payload.get("Value")
                 if dd_json_data:
-                    dd_json_data = base64.b64decode(dd_json_data)
-            elif dd_json_data_type == "String":
-                dd_json_data = dd_payload.get("stringValue") or dd_payload.get("Value")
+                    dd_data = json.loads(dd_json_data)
+
+                    if is_step_function_event(dd_data):
```
michael-zhao459, line 306: DSM does not support Step Functions, I think it's safe to return early here

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                        context = propagator.extract(dd_data)
+                    dsm_data = dd_data
             else:
-                logger.debug(
-                    "Datadog Lambda Python only supports extracting trace"
-                    "context from String or Binary SQS/SNS message attributes"
-                )
+                # Handle case where trace context is injected into attributes.AWSTraceHeader
+                # example:Root=1-654321ab-000000001234567890abcdef;Parent=0123456789abcdef;Sampled=1
+                attrs = event.get("Records")[0].get("attributes")
```
piochelepiotr: here, it's accessing records[0] again. I would put this whole section in an `else if idx == 0` (line 315)

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                        try:
+                            return extract_context_from_step_functions(dd_data, None)
+                        except Exception:
+                            logger.debug(
+                                "Failed to extract Step Functions context from SQS/SNS event."
+                            )
+                    if idx == 0:
+                        context = propagator.extract(dd_data)
+                    dsm_data = dd_data
             else:
```
piochelepiotr: in this else, dsm_data is not set. Is that an issue?

datadog_lambda/tracing.py
```diff
@@ -393,30 +402,42 @@ def extract_context_from_kinesis_event(event, lambda_context):
...
-    _dsm_set_checkpoint(None, "kinesis", source_arn)
-    return extract_context_from_lambda_context(lambda_context)
+    records = (
+        [get_first_record(event)]
+        if not config.data_streams_enabled
+        else event.get("Records")
+    )
+    context = None
+    for idx, record in enumerate(records):
+        dsm_data = None
```
piochelepiotr: this is not specific to dsm, it's `dd_ctx`

datadog_lambda/tracing.py
```diff
@@ -393,30 +402,42 @@ def extract_context_from_kinesis_event(event, lambda_context):
...
+        try:
+            source_arn = record.get("eventSourceARN", "")
+            kinesis = record.get("kinesis")
+            if not kinesis:
+                context = (
+                    extract_context_from_lambda_context(lambda_context)
+                    if idx == 0
+                    else context
+                )
+                _dsm_set_checkpoint(None, "kinesis", source_arn)
```
piochelepiotr: we are setting a checkpoint `kinesis` if not kinesis, I think the name `kinesis` is wrong, because this code is a bit confusing.

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+    context = None
+    records = (
+        event.get("Records", [])
+        if config.data_streams_enabled
+        else [event.get("Records")[0]]
+    )
+    for idx, record in enumerate(records):
+        try:
+            source_arn = record.get("eventSourceARN", "")
+            dsm_data = None
```
piochelepiotr: same as bellow, dsm_data is not specific to data streams here. I would name it same as bellow, `dd_ctx`, or something like that.

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
-        dd_payload = msg_attributes.get("_datadog")
-        if dd_payload:
-            # SQS uses dataType and binaryValue/stringValue
-            # SNS uses Type and Value
-            dd_json_data = None
-            dd_json_data_type = dd_payload.get("Type") or dd_payload.get("dataType")
-            if dd_json_data_type == "Binary":
-                import base64
+    context = None
+    records = (
```
piochelepiotr: Maybe let's have records always be: `event.get("Records", [])`, however, in the loop, we can break early if data streams is disabled.

piochelepiotr:
```
if index == 0:
     do apm stuff
     if data streams is enabled:
         break
```

datadog_lambda/tracing.py
michael-zhao459: Not an issue. I checked in dd-trace-py and DSM never injects context into attributes.AWSTraceHeader, we can just set a checkpoint with None

datadog_lambda/tracing.py
michael-zhao459: hmm. This is deep enough inside the extract function where we believe that the event source is from Kinesis from parsing beforehand. However, all AWS documentation says Kinesis lambda event should have this field. To my understanding, this check is for lambda synchronous invocations with records that match Kinesis, but doesn't actually come from a Kinesis stream. @DataDog/apm-serverless  Can you help confirm why this check is here in the first place?

datadog_lambda/tracing.py
piochelepiotr:
```suggestion
                attrs = record.get("attributes")
```

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                            try:
+                                return extract_context_from_step_functions(
+                                    dd_data, None
+                                )
+                            except Exception:
+                                logger.debug(
+                                    "Failed to extract Step Functions context from SQS/SNS event."
+                                )
+                        context = propagator.extract(dd_data)
+                        if not config.data_streams_enabled:
```
piochelepiotr: this break is too hidden. It is in `if dd_payload` block?

I suggest this high level approach:
```
apm_context = None
for record in records:
  context = extract_context(record)
  if apm_context is None:
      apm_context = context
  if data_streams_enabled:
    set_checkpoint()
  if !data_streams_enabled:
    # APM only looks at the first record.
    break
```

You can break down the code into helper function to make that structure be very clear.
Basically, can you avoid some of the nested conditions? Like `if not config.data_streams_enabled` here?

datadog_lambda/tracing.py
```diff
@@ -248,91 +248,100 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                            break
+                    dd_ctx = dd_data
+            elif idx == 0:
+                # Handle case where trace context is injected into attributes.AWSTraceHeader
+                # example:Root=1-654321ab-000000001234567890abcdef;Parent=0123456789abcdef;Sampled=1
+                attrs = event.get("Records")[0].get("attributes")
+                if attrs:
+                    x_ray_header = attrs.get("AWSTraceHeader")
```
piochelepiotr: maybe put this logic in the `extract_context` I suggested above. The extract_context can take an argument: extract_from_xray?

(extract_context is probably not a great name, I let you find a better one)

michael-zhao459: Hmm. I might be misinterpreting but I'm not sure we should have one function return both a Context() object and the return of a json.loads(). I ended up splitting the x-ray extractor into another helper, let me know what you think

datadog_lambda/tracing.py
```diff
@@ -248,91 +247,105 @@ def extract_context_from_sqs_or_sns_event_or_context(
     except Exception:
         logger.debug("Failed extracting context as EventBridge to SQS.")
 
-    try:
```
piochelepiotr, line 251: if context is extracted from event bridge, we don't set a checkpoint. Is that expected?

datadog_lambda/tracing.py
```diff
@@ -248,91 +247,105 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+            break
+
+    return (
+        apm_context
+        if apm_context
+        else extract_context_from_lambda_context(lambda_context)
+    )
+
+
+def _extract_context_from_sqs_or_sns_record(record):
```
piochelepiotr, line 276: that function looks great to me

datadog_lambda/tracing.py
```diff
@@ -248,91 +247,105 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+        source_arn = (
+            record.get("eventSourceARN")
+            if event_type == "sqs"
+            else record.get("Sns", {}).get("TopicArn")
+        )
+        dd_ctx = None
+        try:
+            dd_ctx = _extract_context_from_sqs_or_sns_record(record)
+            if apm_context is None:
+                if dd_ctx and is_step_function_event(dd_ctx):
```
piochelepiotr: I guess the DSM context can't be in a step_funtion_event?

In any case, the logic to get the apm context from the dd_ctx, should be in it's own function I this.

That function can be the `_extract_context_from_xray` that you can rename to:
`_extract_apm_context`, and it can take the parameters: `dd_ctx` and `record`.

Then, code here can be:
```
if apm_context is None:
       apm_context = _extract_apm_context(dd_ctx, record)
```

piochelepiotr: This will make the code easier to read, but also less error prone.

Here, if the context is extracted from step function, we are not setting a checkpoint. I don't think this is what we want?

datadog_lambda/tracing.py
```diff
@@ -248,91 +247,105 @@ def extract_context_from_sqs_or_sns_event_or_context(
...
+                "context from String or Binary SQS/SNS message attributes"
+            )
+
+        if dd_json_data:
+            dd_data = json.loads(dd_json_data)
+            return dd_data
+    return None
+
+
+def _extract_context_from_xray(record):
```
piochelepiotr: as mentioned above, I would change function to extract the apm context. Not just from xray.

datadog_lambda/tracing.py
michael-zhao459: The tracers never inject DSM context in the case of event bridge or step functions. I'm not sure this is the PR to be adding the functionality for these event types

PR #652, opened 2025-08-20
chore(build): strip debug symbols using `strip -g`

taegyunkim:
### What does this PR do?

Run `strip -g` on all .so files in ddtrace to make sure that they're stripped of debug symbols to reduce the layer size. 

This is needed to unblock https://github.com/DataDog/dd-trace-py/pull/14286, which aims to build dd-trace-py with debug symbols by default. 

The more right thing to do would be using pre-built wheel always instead of building from sources.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,10 +22,10 @@ RUN curl https://sh.rustup.rs -sSf | \
     sh -s -- --default-toolchain stable -y
 ENV PATH=/root/.cargo/bin:$PATH
 
-# Compile mode is explicitly set here to strip the debug symbols from the
-# native extensions in dd-trace-py. Otherwise, they will have debug symbols
-# by default when built from sources. PyPI packages are stripped off of debug
-# symbols. This is mainly to reduce the layer size at the cost of debuggability
+# Compile mode is explicitly set here to not build native extensions with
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,6 +22,9 @@ RUN curl https://sh.rustup.rs -sSf | \
     sh -s -- --default-toolchain stable -y
 ENV PATH=/root/.cargo/bin:$PATH
 
+ENV DD_COMPILE_MODE=Release
+RUN echo "DD_COMPILE_MODE=$DD_COMPILE_MODE"
```
purple4reina: Help me understand, what does this do exactly?

taegyunkim: I'm trying to force dd-trace-py to be built with Release mode, to not include any debug symbols, but it doesn't seem to work.

taegyunkim: OK, setting `ENV DD_COMPILE_MODE=Release` works but I needed to run `strip -g` on all `.so` files to make sure that the binaries indeed don't have debug symbols.

purple4reina: Cool cool. And what happens when you don't add `DD_COMPILE_MODE=Release`? Are the symbols still added?

PR #635, opened 2025-07-17
Enhance Synthetic Span Service Representation

zarirhamza:
### What does this PR do?

Rollout of span naming changes to align serverless product with tracer to create streamlined Service Representation for Serverless

Key Changes:

- Change service name to match instance name for all managed services (aws.lambda -> lambda name, etc) (breaking)

- Add `span.kind:server` on synthetic spans made via span-inferrer

- Implement `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` (default true) to allow users to opt out

### Motivation

Improve Service Map for Serverless. This allows for synthetic spans to have their own service on the map which connects with the inferred spans from the tracer side.

### Testing Guidelines

Unit and Integration tests updated accordingly.

Matching the service name to the instance name essentially removes the need for a base_service tag in some instances due to essentially removing service overrides

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [x] This PR's description is comprehensive
- [x] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [x] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog
- [x] This PR passes the integration tests (ask a Datadog member to run the tests)

PR #654, opened 2025-09-02
[APPSEC]: fix Lambda not appearing on the Appsec API Security Endpoint Catalog

florentinl:
### What does this PR do?

This PR adds the `span.kind` tag with the value `server` for all events that represent an HTTP Request. The api security backend expects it to be present and omits endpoints that do not have this tag. (documentation for this attribute can be found [here](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2357395856/Span+attributes#span.kind))

Additionally API Security should be enabled by default, there is no need to intentionally disable it for Appsec in lambda.

### Motivation

The goal of this PR is to make Lambda services with Appsec API Security enabled show up on the API Security Catalog.

### Testing Guidelines

Updated existing tests

### Additional Notes

- This issue also impacts API Security when it is enabled through the extension.

- As of today, the API Security Endpoint Catalog uses the `resource_name` to display the endpoint and expects it to be in the format `GET /path/to/resource` but lambda sets it to the function name instead. This should maybe be solved in the backend and will be done at a later time.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/__init__.py
+++ b/datadog_lambda/__init__.py
@@ -8,8 +8,6 @@ if os.environ.get("DD_INSTRUMENTATION_TELEMETRY_ENABLED") is None:
         "DD_APPSEC_ENABLED", "false"
     )
 
-if os.environ.get("DD_API_SECURITY_ENABLED") is None:
-    os.environ["DD_API_SECURITY_ENABLED"] = "False"
 
 initialize_cold_start_tracing()
 
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -288,7 +288,7 @@ def extract_http_tags(event):
     """
     Extracts HTTP facet tags from the triggering event
     """
-    http_tags = {}
+    http_tags = {"span.kind": "server"}
 
     # Safely get request_context and ensure it's a dictionary
     request_context = event.get("requestContext")
--- a/tests/integration/snapshots/logs/async-metrics_python310.log
+++ b/tests/integration/snapshots/logs/async-metrics_python310.log
@@ -104,6 +104,7 @@ HTTP GET https://www.datadoghq.com/ Headers: ["Accept-Encoding:gzip, deflate","A
           "span.name": "aws.lambda",
           "function_trigger.event_source": "api-gateway",
           "function_trigger.event_source_arn": "XXXX",
+          "span.kind": "server",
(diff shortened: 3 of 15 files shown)


datadog_lambda/__init__.py
```diff
@@ -8,8 +8,6 @@
         "DD_APPSEC_ENABLED", "false"
     )
 
-if os.environ.get("DD_API_SECURITY_ENABLED") is None:
-    os.environ["DD_API_SECURITY_ENABLED"] = "False"
```
purple4reina, line 12: @florentinl question: what does the `DD_API_SECURITY_ENABLED` env var do? What impact will there be from removing this?

datadog_lambda/trigger.py
```diff
@@ -288,7 +288,7 @@ def extract_http_tags(event):
     """
     Extracts HTTP facet tags from the triggering event
     """
-    http_tags = {}
+    http_tags = {"span.kind": "server"}
```
zarirhamza, line 291: Good catch, I originally made [changes](https://github.com/DataDog/datadog-lambda-python/pull/635) to add this tag to all the related synthetic spans but missed it on the actual `aws.lambda` span. Is there a plan on the roadmap for getting this done in the Node Layer and the Rust extension as well?

datadog_lambda/__init__.py
florentinl: This env controls the enablement of the API Security feature of Appsec. It extracts the schema of the request and runs additional security rules.
On all tracers, it is enabled by default and discussing it with the security ecosystems team there is no reason to change the default for lambda.

Of course, this feature is a sub feature of Appsec and is automatically disabled when DD_APPSEC_ENABLED is not set to a truthy value ([see APIManager enablement in dd-trace-py](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/appsec/_listeners.py#L41))

datadog_lambda/trigger.py
florentinl: Nothing formally planned yet. But we are interested in having it across all lambda layers to fix API Security in lambda (already available through the go and now rust extension).
I can look into it, if you want.

purple4reina (approved):

PR #655, opened 2025-09-02
[APPSEC]: blocking for alb multi value headers events

florentinl:
### What does this PR do?

Currently, ASM returns the same response for all event types. It works in all cases except `ALB` with a target group that has the `multi-value headers` option turned on. In this case, headers must be sent in the `multiValueHeaders` field with type `dict[str, list[str]]` instead of in the `headers` field with type `dict[str, str]`.

This PR fixes blocking for Appsec in the context of ALB events for lambdas in target groups with multi-value headers enabled:

- Add EventSubTypes for regular ALB and ALB multi-value headers to propagate the information
- Send a blocking response with `multiValueHeaders` instead of `headers` when required.
- [typo] renamed the sample events file

### Motivation

While adding the ALB event types to the system-tests in :  DataDog/system-tests#5181. I noticed that blocking responses did not work as expected in the multi-value headers case.

### Testing Guidelines

- updated the unit tests
- the system-tests APPSEC_LAMBDA_BLOCKING scenario asserts that it works as it should:
  - in the [job logs](https://github.com/DataDog/datadog-lambda-python/actions/runs/17546128232/job/49828129072?pr=655#step:93:108) for the alb-multi event type, you can see all `xpassed` tests that were failing currently for blocking and are now passing
- manual testing: spinned up an ALB with multi-value headers and tested it.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


joeyzhao2018 (approved):
LGTM

PR #656, opened 2025-09-08
fix(appsec): perform _after extractions on the blocking_response 

florentinl:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/wrapper.py
```diff
@@ -284,6 +284,30 @@ def _before(self, event, context):
...
+                        response=self.response,
+                    )
+                    self.blocking_response = get_asm_blocked_response(self.event_source)
+
+                if self.blocking_response:
+                    status_code = self.blocking_response.get("statusCode")
+                    self.response = self.blocking_response
+
+                if config.capture_payload_enabled:
+                    tag_object(self.span, "function.request", event)
```
joeyzhao2018, line 307: not a request to change but I wonder if we should still tag the original self.response? maybe like tage it "function.original_request"? Does being blocked means we don't want the customer to see the original response whatsoever?

florentinl: We don't particularly want to hide it from the customer, the question is rather: would it be useful to have it ? The security signal will already have a detail of which security rule was matched to trigger the block. Having the response might be handy for debugging.

The request can be blocked at multiple moments. We would only have an "original response" when we block after looking at the response of the handler. 
In this case, I can tag a "function.blocked_response" to link it more explicitly with the blocking feature.

PR #657, opened 2025-09-12
APMSVLS-65 Extract Trace Context For AppSync Events

rithikanarayan:
### What does this PR do?
This PR adds logic to extract the trace context of an event that comes from a traced service through AppSync to a Python lambda. When a lambda is invoked from an AppSync API which was called in a RUM-instrumented front-end, the datadog trace context is located under event["request"]["headers"] rather than other locations from different event types. The extraction logic is placed directly in the `extract_dd_trace_context` function in tracing.py, indexing into the request key in the event, after which the `extract_context_from_http_event_or_context` function will get the header field and extract datadog context properly. 

### Motivation
Currently, if a customer has a setup where they use RUM to start a trace that goes through AppSync and triggers a Lambda function, we ask them to write a custom function to extract the trace context from the AppSync event that is passed to the Lambda. See an example of such a function [here.](https://github.com/joeyzhao2018/my_appsync_cdk/blob/main/lambda/layers/common/python/custom_extractor.py) It would be simpler for customers and for ourselves to extract the trace context in the lambda tracer layers ourselves, as we do for invocations from other sources (ex. SQS, API Gateway, etc.). This PR does so for the Python tracer. 

Based on [this ticket](https://datadoghq.atlassian.net/browse/APMSVLS-65), will reduce amount of code changes a customer has to make in order connect traces between RUM and a Lambda function when there is an AWS AppSync API in between them. 

### Testing Guidelines

Unit tested in `test_tracing.py` by adding events in tests/event_samples and including the tests in _test_extract_dd_trace_context. The events added, `rum-appsync.json`, `rum-appsync-no-headers.json`, and `rum-appsync-request-not-dict.json`, are based on a sample request from the Datadog APM page of a trace that followed RUM -> AppSync -> Lambda. Some of the sample events are malformed/formatted differently than expected to ensure that exceptions are not raised if we encounter an event with a different format than anticipated. I ran a coverage test using pytest-cov to ensure that all new lines of code from this PR were tested. 

Ran integration tests using scripts/run_integration_tests.sh. Added a new input event called `appsync.json` and updated the snapshot so that integration tests also cover this new supported case. 

Uploaded my changes as a layer to AWS and tested whether a trace that followed RUM -> AppSync -> Lambda was shown as connected in the Datadog UI without needing a custom extractor, which is the goal of this PR. A successfully connected trace can be found [here](https://app.datadoghq.com/apm/traces?agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&sort_by=resource_name&sort_order=desc&spanID=8623521558131511024&spanType=all&spanViewType=metadata&storage=hot&timeHint=1758654385966&trace=102112735132931059768623521558131511024&traceID=10211273513293105976&traceQuery=&view=spans&start=1758653494048&end=1758654394048&paused=false). The ARN for the testing version of the Python Lambda layer is `arn:aws:lambda:us-east-1:425362996713:layer:Python39-RITHIKA:3` Also used this layer to check distributed tracing in an API Gateway -> Lambda -> SQS -> Lambda setup to ensure that other tracing functionality was not broken by change. 

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR's description is comprehensive
- [X] This PR passes the integration tests (ask a Datadog member to run the tests)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - "main"
   pull_request:
+  schedule:
+    - cron: '0 0,12 * * *' # Runs every day at midnight and noon utc
 
 jobs:
   lint:
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -46,6 +46,10 @@ from datadog_lambda.trigger import (
     extract_http_status_code_tag,
 )
 
+# ddtrace imports are also tested in
+# dd-trace-py/tests/internal/test_serverless.py please update those tests when
+# making changes to any ddtrace import.
+
 if config.appsec_enabled:
     from datadog_lambda.asm import (
         asm_set_context,
--- a/scripts/check_layer_size.sh
+++ b/scripts/check_layer_size.sh
@@ -8,7 +8,7 @@
 # Compares layer size to threshold, and fails if below that threshold
 
 set -e
-MAX_LAYER_COMPRESSED_SIZE_KB=$(expr 9 \* 1024)  #  9216 KB
+MAX_LAYER_COMPRESSED_SIZE_KB=$(expr 9 \* 1024)      # 9216 KB
 MAX_LAYER_UNCOMPRESSED_SIZE_KB=$(expr 25 \* 1024)   # 25600 KB
 
 
--- a/tests/test_wrapper.py
+++ b/tests/test_wrapper.py
@@ -789,3 +789,113 @@ class TestLambdaWrapperAppsecBlocking(unittest.TestCase):
         self.mock_asm_start_response.assert_not_called()
 
         assert lambda_handler.span.get_tag("http.status_code") == "200"
+
+
+@patch("datadog_lambda.config.Config.exception_replay_enabled", True)
+def test_exception_replay_enabled(monkeypatch):
+    importlib.reload(wrapper)
+
+    original_SpanExceptionHandler_enable = wrapper.SpanExceptionHandler.enable
+    SpanExceptionHandler_enable_calls = []
+
+    def SpanExceptionHandler_enable(*args, **kwargs):
+        SpanExceptionHandler_enable_calls.append((args, kwargs))
+        return original_SpanExceptionHandler_enable(*args, **kwargs)
+
+    original_SignalUploader_periodic = wrapper.SignalUploader.periodic
+    SignalUploader_periodic_calls = []
+
+    def SignalUploader_periodic(*args, **kwargs):
+        SignalUploader_periodic_calls.append((args, kwargs))
+        return original_SignalUploader_periodic(*args, **kwargs)
+
+    monkeypatch.setattr(
+        "datadog_lambda.wrapper.SpanExceptionHandler.enable",
+        SpanExceptionHandler_enable,
+    )
+    monkeypatch.setattr(
+        "datadog_lambda.wrapper.SignalUploader.periodic", SignalUploader_periodic
+    )
+
+    expected_response = {
+        "statusCode": 200,
+        "body": "This should be returned",
+    }
+
+    @wrapper.datadog_lambda_wrapper
+    def lambda_handler(event, context):
+        return expected_response
+
+    response = lambda_handler({}, get_mock_context())
+
+    assert response == expected_response
+    assert len(SpanExceptionHandler_enable_calls) == 1
+    assert len(SignalUploader_periodic_calls) == 1
+
+
+@patch("datadog_lambda.config.Config.profiling_enabled", True)
+def test_profiling_enabled(monkeypatch):
+    importlib.reload(wrapper)
+
+    original_Profiler_start = wrapper.profiler.Profiler.start
+    Profiler_start_calls = []
+
+    def Profiler_start(*args, **kwargs):
+        Profiler_start_calls.append((args, kwargs))
+        return original_Profiler_start(*args, **kwargs)
+
+    monkeypatch.setattr("datadog_lambda.wrapper.is_new_sandbox", lambda: True)
+    monkeypatch.setattr(
+        "datadog_lambda.wrapper.profiler.Profiler.start", Profiler_start
+    )
+
+    expected_response = {
+        "statusCode": 200,
+        "body": "This should be returned",
+    }
+
+    @wrapper.datadog_lambda_wrapper
+    def lambda_handler(event, context):
+        return expected_response
+
+    response = lambda_handler({}, get_mock_context())
+
+    assert response == expected_response
+    assert len(Profiler_start_calls) == 1
+
+
+@patch("datadog_lambda.config.Config.llmobs_enabled", True)
+def test_llmobs_enabled(monkeypatch):
+    importlib.reload(wrapper)
+
+    original_LLMObs_enable = wrapper.LLMObs.enable
+    LLMObs_enable_calls = []
+
+    def LLMObs_enable(*args, **kwargs):
+        LLMObs_enable_calls.append((args, kwargs))
+        return original_LLMObs_enable(*args, **kwargs)
+
+    original_LLMObs_flush = wrapper.LLMObs.flush
+    LLMObs_flush_calls = []
+
+    def LLMObs_flush(*args, **kwargs):
+        LLMObs_flush_calls.append((args, kwargs))
+        return original_LLMObs_flush(*args, **kwargs)
+
+    monkeypatch.setattr("datadog_lambda.wrapper.LLMObs.enable", LLMObs_enable)
+    monkeypatch.setattr("datadog_lambda.wrapper.LLMObs.flush", LLMObs_flush)
+
+    expected_response = {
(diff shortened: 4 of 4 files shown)


datadog_lambda/tracing.py
```diff
@@ -536,6 +536,25 @@ def extract_context_from_step_functions(event, lambda_context):
         logger.debug("The Step Functions trace extractor returned with error %s", e)
         return extract_context_from_lambda_context(lambda_context)
 
+def extract_context_from_appsync_event(event, lambda_context):
```
joeyzhao2018: I wonder if this is only for appsync

datadog_lambda/tracing.py
```diff
@@ -536,6 +536,25 @@ def extract_context_from_step_functions(event, lambda_context):
...
+    """
+    Extract Datadog trace context from an AppSync event. Intended to replace
+    need for customers to provide custom extractor function in their code. 
+    """
+    nested_json = event.get("request", {}).get("headers", {})
+    trace_id = nested_json.get("x-datadog-trace-id")
```
joeyzhao2018: we should delegate the extraction to the ddtrace extractor

datadog_lambda/tracing.py
```diff
@@ -641,6 +660,8 @@ def extract_dd_trace_context(
         context = extract_context_from_kinesis_event(event, lambda_context)
     elif event_source.equals(EventTypes.STEPFUNCTIONS):
         context = extract_context_from_step_functions(event, lambda_context)
+    elif event_source.equals(EventTypes.APPSYNC):
```
joeyzhao2018: note: not reachable as we don't have the `EventTypes.APPSYNC`

datadog_lambda/tracing.py
```diff
@@ -537,6 +537,25 @@ def extract_context_from_step_functions(event, lambda_context):
         return extract_context_from_lambda_context(lambda_context)
 
 
+def extract_context_from_request_header(event, lambda_context):
+    """
+    Attempt to extract Datadog trace context from an event payload where Datadog information
+    is under the event["request"]["header"] section. Intended to replace some of the
+    need for customers to provide custom extractor functions in their code for workflows such as
+    RUM -> AppSync -> Datadog.
```
purple4reina:
```suggestion
    RUM -> AppSync -> Lambda.
```

datadog_lambda/tracing.py
```diff
@@ -537,6 +537,25 @@ def extract_context_from_step_functions(event, lambda_context):
...
+    """
+    try:
+        nested_json = event.get("request", {}).get("headers", {})
```
purple4reina: In the case where `"request"` is found in the event dict, the extra `{}` is unnecessary. Allocations are the slowest thing in python, so we want to avoid them as much as possible in this repo.

There's a lot of ways to get around this. You could `event.get("request") or {}`, you could `event.get("request")` then check if the returned value is `None`.

rithikanarayan: I'm removing the allocations but not using either of the fixes you mentioned as of yet, just handling it with the try except instead of nested ifs for the None checks. I think the `or {}` fix is unnecessary because we don't need the key to exist if it doesn't, we are just looking for its contents if it exists. Let me know if this isn't good practice!

purple4reina: Using the try/except logic does work. But I find that when an exception is raised, it actually takes a _long_ time for python to do its thing and send you into the except block.

Have you seen/used the `timeit` python module yet? I'd recommend exploring your solution using benchmarking.

datadog_lambda/tracing.py
```diff
@@ -537,6 +537,25 @@ def extract_context_from_step_functions(event, lambda_context):
...
+    try:
+        nested_json = event.get("request").get("headers")
+        context = propagator.extract(nested_json)
+
+        if not _is_context_complete(context):
+            context = extract_context_from_lambda_context(lambda_context)
+    except Exception:
+        context = extract_context_from_lambda_context(lambda_context)
+
+    return context
```
purple4reina:
```suggestion
        if _is_context_complete(context):
            return context
    except Exception:
        pass

    return extract_context_from_lambda_context(lambda_context)
```

rithikanarayan: Discussed refactoring change to use existing extract_context_from_http_event_or_context, suggested change no longer needed.

datadog_lambda/tracing.py
```diff
@@ -627,6 +627,14 @@ def extract_dd_trace_context(
 
     if extractor is not None:
         context = extract_context_custom_extractor(extractor, event, lambda_context)
+    elif isinstance(event, (set, dict)) and "request" in event:
+        if "headers" in event.get("request"):
+            context = extract_context_from_http_event_or_context(
+                event.get("request"),
+                lambda_context,
+                event_source,
+                decode_authorizer_context=False,
```
joeyzhao2018: why not using the given `decode_authorizer_context` here?

rithikanarayan: From my understanding of the code, if decode_authorizer_context was set to True, we would then call the function get_injected_authorizer_data (within extract_context_from_http_event_or_context)and check for fields like requestContext, which we don't expect to see and get_injected_authorized_data would return None. We would then try to get the headers from the event and extract context from those anyway, so it seemed unnecessary to enter the first "if decode_authorizer_context" code block in extract_context_from_http_event_or_context. If decode_authorizer_context was set to True, there's a possibility that this PR would be more generalizable, but I don't actually know whether there are scenarios where the context is in event["request"]["requestContext"] and the like.

datadog_lambda/tracing.py
```diff
@@ -627,6 +627,14 @@ def extract_dd_trace_context(
 
     if extractor is not None:
         context = extract_context_custom_extractor(extractor, event, lambda_context)
+    elif isinstance(event, (set, dict)) and "request" in event:
+        if "headers" in event.get("request"):
+            context = extract_context_from_http_event_or_context(
+                event.get("request"),
```
purple4reina: You're calling `event.get("request")` twice. Why not only do it once?

datadog_lambda/tracing.py
joeyzhao2018: Thanks for the detailed explanations.

datadog_lambda/tracing.py
purple4reina: If a event like this is used, then this code will raise an exception:

```json
{
  "request": "hello world"
}
```

rithikanarayan: Addressed in following commits!

datadog_lambda/tracing.py
```diff
@@ -627,6 +627,17 @@ def extract_dd_trace_context(
...
+        request = event.get("request")
+        if isinstance(request, (set, dict)) and "headers" in request:
+            context = extract_context_from_http_event_or_context(
+                request,
+                lambda_context,
+                event_source,
+                decode_authorizer_context=False,
+            )
+        else:
+            context = extract_context_from_lambda_context(lambda_context)
```
purple4reina: Nit (meaning you don't need to make this change if you don't want to)

There's now enough here to warrant moving this logic into its own method. That would keep the code clean and make unit testing easier.

tests/test_tracing.py
```diff
@@ -179,6 +179,14 @@ def _wrap(*args, **kwargs):
...
     ),
     ("lambda-url", None),
+    (
+        "rum-appsync",
+        Context(
+            trace_id=12345,
+            span_id=67890,
+            sampling_priority=1,
+        ),
+    ),
```
purple4reina, line 189: There's a lot more testing that we're going to need. When thinking about test coverage, I think about two things.

1. Coverage. If I were to run a coverage report on our tests, would I see that we've hit every line of code? In the current case, the answer is no.

2. Logic. If I were to make a change to any line of your code, would a test fail? For example, if I change a `in` to `not in`, I would expect to see a failing test. In the current case, this could mean testing that `request` is a dict but `headers` is not in it and vice versa.

We also would need a test that makes sure that the authorizer context is never decoded.

datadog_lambda/tracing.py
```diff
@@ -209,6 +209,20 @@ def extract_context_from_http_event_or_context(
...
+    request = event.get("request")
+    if isinstance(request, (set, dict)) and "headers" in request:
+        context = extract_context_from_http_event_or_context(
+            request,
+            lambda_context,
+            event_source,
+            decode_authorizer_context=False,
+        )
+    else:
+        context = extract_context_from_lambda_context(lambda_context)
```
purple4reina, line 222: Looks like we still need a test to cover this portion of the logic. We know that the function does not error when the request is not a dict or that there are no headers in it. But we have not confirmed that we'll instead attempt to extract from the lambda context.

PR #663, opened 2025-10-09
fix(http): http.url tag should contain the path

florentinl:
### What does this PR do?

The `http.url_details.path` tag is set by the backend as described in [the span attributes guide of APM](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2357395856/Span+attributes#http.url_details.path). When set by this tracer, it is replaced by the backend by a value parsed from the `http.url` which is always `/`.

Here is an example of a trace received from an API Gateway event where the route can be correctly determined but the path is wrongly set for this reason:
image

The path should be appended to the `http.url` tag instead.

### Motivation

- In the App and API Protection security traces explorer, we show the endpoint and the path is reported incorrectly for this reason.
image 

instead of the expected:
image

### Testing Guidelines

- Updated snapshots
- Updated unit tests
- Tested with API Gateway events where scheme and host are extracted
- Tested with ALB events where `http.url` is only set to the path

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/trigger.py
+++ b/datadog_lambda/trigger.py
@@ -325,7 +325,8 @@ def extract_http_tags(event):
             method = apigateway_v2_http.get("method")
 
     if path:
-        http_tags["http.url_details.path"] = path
+        if http_tags.get("http.url"):
+            http_tags["http.url"] += path
     if method:
         http_tags["http.method"] = method
 
--- a/tests/integration/snapshots/logs/async-metrics_python310.log
+++ b/tests/integration/snapshots/logs/async-metrics_python310.log
@@ -105,8 +105,7 @@ HTTP GET https://www.datadoghq.com/ Headers: ["Accept-Encoding:gzip, deflate","A
           "function_trigger.event_source": "api-gateway",
           "function_trigger.event_source_arn": "XXXX",
           "span.kind": "server",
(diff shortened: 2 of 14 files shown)


datadog_lambda/trigger.py
```diff
@@ -325,7 +325,10 @@ def extract_http_tags(event):
             method = apigateway_v2_http.get("method")
 
     if path:
-        http_tags["http.url_details.path"] = path
+        if http_tags.get("http.url"):
+            http_tags["http.url"] += path
+        else:
```
zarirhamza: Does this fallback make sense? Do we want the `url` to be just the path if the preceding `scheme://host` cannot be found?

Not sure what the backend logic is here just want to confirm

florentinl: The spec isn't very clear on if it is intended or not but I could test it with an ALB (case where we have no url) and it works as you would expect.
The backend correctly deduces the `http.url_details.path` and it is also now deduce the `http.pathgroup` tag as well.

zarirhamza: Yeah lets just make sure the URL tag _only_ has URL since other teams rely on it

PR #672, opened 2025-10-29
feat(config): enable telemetry when SCA is on independently from appsec

florentinl:
### What does this PR do?

My understanding is that we want instrumentation telemetry to be off by default in python lambda. There is currently a bypass for AppSec that is a bit weird using env vars directly in `__init__.py`. I added another bypass for Runtime SCA and moved the logic to the `config` module to rely on parsed env vars instead.

### Motivation

A customer asked if he could enable Runtime SCA without enabling AAP (AppSec) in Lambda. This currently requires setting `DD_INSTRUMENTATION_TELEMETRY_ENABLED` manually, we could make it a bit more easy.

### Additional Notes

- In `ddtrace`, the telemetry_writer is instantiated and enabled on first import if `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is not explicitely set to false => The current telemetry enablement code in `datadog_lambda/tracing.py` is therefore redundant and unnecessary so I removed it.
- The `config.enable_telemetry` flag was therefore not necessary anymore and I removed it.

- This small refactor relies on the fact that the configuration is resolved before importing ddtrace which happens transitively through `cold_start.py`. Should we make this more explicit by importing it in `__init__.py` or is this already clear enough ? 

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/datadog_lambda/__init__.py
+++ b/datadog_lambda/__init__.py
@@ -1,12 +1,5 @@
+import datadog_lambda.config  # noqa: F401 needs to be imported before `ddtrace`
 from datadog_lambda.cold_start import initialize_cold_start_tracing
-import os
-
-
-if os.environ.get("DD_INSTRUMENTATION_TELEMETRY_ENABLED") is None:
-    # Telemetry is required for Appsec Software Composition Analysis
-    os.environ["DD_INSTRUMENTATION_TELEMETRY_ENABLED"] = os.environ.get(
-        "DD_APPSEC_ENABLED", "false"
-    )
 
 
 initialize_cold_start_tracing()
--- a/datadog_lambda/config.py
+++ b/datadog_lambda/config.py
@@ -82,12 +82,6 @@ class Config:
     logs_injection = _get_env("DD_LOGS_INJECTION", "true", as_bool)
     merge_xray_traces = _get_env("DD_MERGE_XRAY_TRACES", "false", as_bool)
 
-    telemetry_enabled = _get_env(
-        "DD_INSTRUMENTATION_TELEMETRY_ENABLED",
-        "false",
-        as_bool,
-        depends_on_tracing=True,
-    )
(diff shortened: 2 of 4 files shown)


datadog_lambda/__init__.py
```diff
@@ -1,12 +1,4 @@
 from datadog_lambda.cold_start import initialize_cold_start_tracing
-import os
-
-
-if os.environ.get("DD_INSTRUMENTATION_TELEMETRY_ENABLED") is None:
-    # Telemetry is required for Appsec Software Composition Analysis
-    os.environ["DD_INSTRUMENTATION_TELEMETRY_ENABLED"] = os.environ.get(
-        "DD_APPSEC_ENABLED", "false"
-    )
```
purple4reina, line 9: The reason this code was added here was we must ensure the `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is set _before_ any ddtrace code is imported. Can we ensure that we the change you're proposing? How can we enforce it?

florentinl: > The reason this code was added here was we must ensure the DD_INSTRUMENTATION_TELEMETRY_ENABLED is set before any ddtrace code is imported.

I agree. As explained in the PR notes, this is the reason why we need the config to be resolved before the `from datadog_lambda.patch import patch_all` line of `__init__.py`.
 
> Can we ensure that we the change you're proposing? How can we enforce it?

We can eagerly import the config in `__init__.py` and add a test to ensure that importing `config.py` does not import `ddtrace`. Would that work for you or does it feel like an unnecessary constraint ? Let me suggest something.

The idea was to reuse the parsing of environment variables done in config if possible. This also looked like some configuration and it felt natural to put it in "config.py".
I can keep it in `__init__.py` if it make more sense.

purple4reina: I think any of those options are fine. As long as we can ensure it's being set before importing ddtrace. I like your idea of adding a test to check for this.

PR #701, opened 2025-12-16
Support ddtrace v4

rithikanarayan:
### What does this PR do?
Conditions ddtrace version based on runtime; Python 3.8 and 3.9 layers will use ddtrace versions >=3.19.1 and <4, while Python 3.10 and above will use ddtrace versions >=4 and <5. 

### Motivation
ddtrace released v4 a few weeks ago with breaking changes and many bug fixes; this PR aligns the lambda layer with that release. 

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,9 +6,6 @@ authors = ["Datadog, Inc. <dev@datadoghq.com>"]
 
 [project]
 name = "datadog_lambda"
-version = "8.121.0.dev0"
-description = "The Datadog AWS Lambda Library"
-authors = [{ name = "Datadog, Inc.", email = "dev@datadoghq.com" }]
 license = "Apache-2.0"
 readme = "README.md"
 requires-python = ">=3.8.0,<4"
@@ -40,6 +37,8 @@ classifiers = [
     "Programming Language :: Python :: 3.14",
 ]
 
+# can remove tool.poetry.depencies and tool.poetry.extras
+# after deprecation of Python 3.8
 [tool.poetry.dependencies]
 python = ">=3.8.0,<4"
 datadog = ">=0.51.0,<1.0.0"


purple4reina:
Whoa, looks like the python 3.14 profiling e2e tests are now passing!

rithikanarayan:
Will need to restrict to >v4.1.1 following that release

pyproject.toml
purple4reina, line 52: So, looks like these dependencies are now listed in two places. I feel like we shouldn't confuse our future selves by having two locations we'd need to update all the time. I don't have much experience with these pyproject.toml files but I'm thinking we can probably delete this entire section?

pyproject.toml
```diff
@@ -46,6 +61,15 @@ dev = [
     "requests",
 ]
```
purple4reina, line 61: And maybe we can delete this section too?

pyproject.toml
```diff
@@ -3,19 +3,34 @@ name = "datadog_lambda"
 version = "8.121.0.dev0"
 description = "The Datadog AWS Lambda Library"
 authors = ["Datadog, Inc. <dev@datadoghq.com>"]
+
+[project]
+name = "datadog_lambda"
+version = "8.121.0.dev0"
```
purple4reina: Looks like we now have the version declared in two places. Do we need both? I'm wondering if we can potentially remove the `tools.poetry` section above.

pyproject.toml
rithikanarayan: Duplicated sections are required while we still support Python 3.8. Tool.poetry.depencies and tool.poetry.extras sections can be deleted following 3.8's deprecation.

pyproject.toml
rithikanarayan: See above. Added comments in file to explain this.

PR #743, opened 2026-03-10
feat: emit durable execution log to stdout at invocation start [SVLS-8582]

lym953:
## Summary

At the beginning of an invocation of a durable function, emit a log like:

```json
{
  "request_id": "123",
  "durable_execution_name": "abc",
  "durable_execution_id": "def",
  "schema_version": "1.0.0"
}
```

The Lambda extension layer uses this log to map request ID to durable_execution_name and durable_execution_id.

Passing this information using logs instead of trace tags so the architecture is cleaner on extension side: the logs agent can use logs to extract this information and add attribute to other logs, instead of depending on traces and receiving information from trace agent. Discussion: https://github.com/DataDog/datadog-lambda-extension/pull/1053#discussion_r2902121545

## Test plan
1. Passed the added unit tests
2. Build a layer, install it on a function, and invoke the function. The message was logged as expected:

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/datadog_lambda/durable.py
+++ b/datadog_lambda/durable.py
@@ -2,10 +2,16 @@
 # under the Apache License Version 2.0.
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
 # Copyright 2019 Datadog, Inc.
+import json
 import logging
 import re
+import sys
 
 logger = logging.getLogger(__name__)
+
+# When changing the schema of the durable invocation log, bump this version so
+# the extension can handle compatibility between old and new schemas.
+DURABLE_INVOCATION_LOG_SCHEMA_VERSION = "1.0.0"
 
 
 def _parse_durable_execution_arn(arn):
@@ -47,3 +53,18 @@ def extract_durable_function_tags(event):
         "durable_function_execution_name": execution_name,
         "durable_function_execution_id": execution_id,
     }
+
+
+def emit_durable_execution_log(request_id, execution_name, execution_id):
+    """
+    Emits a structured JSON log to stdout mapping the Lambda request_id to the
+    durable execution name and ID. This is consumed by the Lambda extension layer
+    to correlate request IDs with durable executions.
+    """
+    log = {
+        "request_id": request_id,
+        "durable_execution_name": execution_name,
+        "durable_execution_id": execution_id,
+        "schema_version": DURABLE_INVOCATION_LOG_SCHEMA_VERSION,
+    }
+    print(json.dumps(log), file=sys.stdout, flush=True)
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -42,7 +42,10 @@ from datadog_lambda.tracing import (
     tracer,
     propagator,
 )
-from datadog_lambda.durable import extract_durable_function_tags, emit_durable_execution_log
+from datadog_lambda.durable import (
+    extract_durable_function_tags,
+    emit_durable_execution_log,
+)
 from datadog_lambda.trigger import (
     extract_trigger_tags,
     extract_http_status_code_tag,
--- a/datadog_lambda/wrapper.py
+++ b/datadog_lambda/wrapper.py
@@ -42,7 +42,7 @@ from datadog_lambda.tracing import (
     tracer,
     propagator,
 )
-from datadog_lambda.durable import extract_durable_function_tags
+from datadog_lambda.durable import extract_durable_function_tags, emit_durable_execution_log
 from datadog_lambda.trigger import (
     extract_trigger_tags,
     extract_http_status_code_tag,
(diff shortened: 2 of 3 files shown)


datadog_lambda/durable.py
```diff
@@ -47,3 +53,18 @@ def extract_durable_function_tags(event):
...
+    durable execution name and ID. This is consumed by the Lambda extension layer
+    to correlate request IDs with durable executions.
+    """
+    log = {
+        "request_id": request_id,
+        "durable_execution_name": execution_name,
+        "durable_execution_id": execution_id,
+        "schema_version": DURABLE_INVOCATION_LOG_SCHEMA_VERSION,
+    }
+    print(json.dumps(log), file=sys.stdout, flush=True)
```
joeyzhao2018, line 70: why not using the logger?

datadog_lambda/durable.py
duncanista, line 1: Although your change says "to use logs instead of traces to enrich logs" still creates a dependency where you need a tracing layer to have complete observability in logs – does that even make sense from a dependency standpoint?

datadog_lambda/durable.py
duncanista: @joeyzhao2018 probably logger injects other things which we don't want

datadog_lambda/durable.py
```diff
@@ -47,3 +53,18 @@ def extract_durable_function_tags(event):
...
+
+
+def emit_durable_execution_log(request_id, execution_name, execution_id):
+    """
+    Emits a structured JSON log to stdout mapping the Lambda request_id to the
+    durable execution name and ID. This is consumed by the Lambda extension layer
+    to correlate request IDs with durable executions.
+    """
+    log = {
+        "request_id": request_id,
```
joeyzhao2018, line 65: I thought extension side has the information of the request_id

datadog_lambda/durable.py
lym953: I think this message should always be logged, and using logger introduces a risk that this message is dropped due to log level. Do you think it's safe to use `logger.info()`?

datadog_lambda/durable.py
```diff
@@ -47,3 +53,18 @@ def extract_durable_function_tags(event):
...
+        "durable_execution_name": execution_name,
+        "durable_execution_id": execution_id,
+        "schema_version": DURABLE_INVOCATION_LOG_SCHEMA_VERSION,
```
joeyzhao2018, line 68: why we need a schema version here?

datadog_lambda/durable.py
joeyzhao2018: ah, you mean the "parsing" part would be a big unpredictable? can we use regex to solve that part? 
using logger can make sure it still follows the existing rules such as extra injections or if customers want to change the level.

datadog_lambda/durable.py
lym953: The dependency refers to the one between trace agent and log agent in lambda extension. Using traces adds coupling between them, making extension code harder to maintain. See discussion here https://github.com/DataDog/datadog-lambda-extension/pull/1053#discussion_r2902121545

datadog_lambda/durable.py
lym953: Race conditions often happen (e.g. logs can be delayed), and given a log message, it's hard for the extension to know which request_id it was for, so I think it's safer and easier to add request_id to the log message.

datadog_lambda/durable.py
lym953: I'm worried that we may need to add more fields in the future, though I'm not sure how yet. Adding schema version will make it easier for extension to identify incompatibility (e.g. missing field) and print an error to ask the user to update tracer.

datadog_lambda/durable.py
joeyzhao2018: oh, i see, you don't want this to be controlled by log level.

joeyzhao2018:
> the logs agent can use logs to extract this information and add attribute to other logs
1. Would there be any race conditions such that this piece of log gets processed too late?
2. Would there be any performance impact since you need to look for one specific log?

datadog_lambda/durable.py
joeyzhao2018: adding more fields shouldn't require a schema. i'm not very convinced that we need it.

datadog_lambda/durable.py
joeyzhao2018: i see. that makes sense.

datadog_lambda/durable.py
jchrostek-dd: I would agree that we don't need schema version. If we need to change this in the future, which I think is very unlikely, then we can add schema version at that point.

lym953:
Closing. We decided to pass the metadata to extension as span tags.

PR #788, opened 2026-04-09
chore(ci): use dd-sts for system-tests test optimization

nccatoni:
## Summary

Migrates system-tests CI to use [dd-sts](https://github.com/DataDog/dd-sts-action) for Datadog Test Optimization instead of long-lived API keys.

All repositories now share a single `system-tests` policy (see [dd-source#408172](https://github.com/DataDog/dd-source/pull/408172)) — no per-repo policy is needed.

Depends on [DataDog/system-tests#6726](https://github.com/DataDog/system-tests/pull/6726).

### Changes
- Add `id-token: write` permission to the system-tests reusable workflow call so it can obtain short-lived credentials via OIDC
- Pin system-tests to `1e5d6b709` (current `main`, pre-migration) to allow a controlled rollout: repos stay on the pre-migration workflow until their pin is explicitly updated to the post-merge SHA

## How to review

The only functional change is adding `id-token: write`. It has no effect at the pinned SHA (dd-sts is not yet used there) but will be required once each repo's pin is updated after [DataDog/system-tests#6726](https://github.com/DataDog/system-tests/pull/6726) merges.

### DIFF
--- a/.github/workflows/system_tests.yml
+++ b/.github/workflows/system_tests.yml
@@ -34,6 +34,7 @@ jobs:
       TEST_OPTIMIZATION_API_KEY: ${{ secrets.TEST_OPTIMIZATION_API_KEY }}
     permissions:
       contents: read
+      id-token: write
       packages: write
     with:
       library: python_lambda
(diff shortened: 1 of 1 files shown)


rithikanarayan (approved):
Makes sense to me. Will there be a follow up PR to update the SHA once the other PR mentioned is merged?

nccatoni:
@rithikanarayan Yes there will be another PR to go back to the ref that was previously used (here main)

PR #809, opened 2026-04-24
enable instrumentation telemetry

emmettbutler:
This change removes the configuration that disables instrumentation telemetry and replaces it with a single telemetry upload during cold start, per the Instrumentation Telemetry For Serverless RFC.

https://github.com/DataDog/dd-trace-py/pull/17661 makes this change work as described in the Instrumentation Telemetry For Serverless RFC.

### DIFF
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -31,6 +31,7 @@ from datadog_lambda.xray import (
 
 from ddtrace import patch
 from ddtrace import __version__ as ddtrace_version
+from ddtrace.internal.telemetry import telemetry_writer
 from ddtrace.propagation.http import HTTPPropagator
 from ddtrace.trace import Context, Span, tracer
 
@@ -1509,6 +1510,10 @@ def create_function_execution_span(
                 pointer_hash=span_pointer_description.pointer_hash,
                 extra_attributes=span_pointer_description.extra_attributes,
             )
+
+    if is_cold_start():
+        telemetry_writer.periodic(force_flush=True)
+
     return span
 
 
--- a/datadog_lambda/tracing.py
+++ b/datadog_lambda/tracing.py
@@ -1511,7 +1511,7 @@ def create_function_execution_span(
                 extra_attributes=span_pointer_description.extra_attributes,
             )
 
-    if is_cold_start():
+    if is_cold_start:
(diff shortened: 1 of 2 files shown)


datadog_lambda/tracing.py
```diff
@@ -31,6 +31,7 @@
 
 from ddtrace import patch
 from ddtrace import __version__ as ddtrace_version
+from ddtrace.internal.telemetry import telemetry_writer
```
purple4reina, line 34: I recall that importing the telemetry writer added a bunch of cold start time. Did you happen to check to see what impact it's had this time around?

emmettbutler: I haven't, though the impact is expected to be reduced by the changes from the RFC. Is there a benchmark somewhere in CI that you'd look at to validate this?

purple4reina: It doesn't look like we have any on this repo, at least none that I can easily find. Is this the only change required to enable instrumentation telemetry? I can get a manual test going.

emmettbutler: Yes I believe it is

purple4reina:
This PR does increase cold start time by approx 4.79% (see first column in chart below).

I deployed a lambda layer from this pr branch and another layer from current on main, then used https://github.com/purple4reina/dd-python-cold-start-testing for testing.  It deploys two lambda functions then invokes them with frequent cold starts.  Results are sent to dashboard https://ddserverless.datadoghq.com/s/61ba5v0dnxcpd8dm/3gu-sxi-fq2.

@emmettbutler you said you're expecting to see some improvements later.  Can you point me to that RFC you referenced and the performance gains we're hoping to see?

Screenshot 2026-04-27 at 13 14 06

emmettbutler:
@purple4reina here's the ddtrace branch that this PR assumes / depends on for the performance gains: https://github.com/DataDog/dd-trace-py/pull/17661

purple4reina:
> @purple4reina here's the ddtrace branch that this PR assumes / depends on for the performance gains: [DataDog/dd-trace-py#17661](https://github.com/DataDog/dd-trace-py/pull/17661)

This PR doesn't look like it's going to improve cold start times at all, which is what the benchmarks are showing we need to address.

PR #818, opened 2026-05-06
durable: extract trace context from checkpoints

joeyzhao2018:
## Summary
https://datadoghq.atlassian.net/browse/APMSVLS-493

Adds Datadog trace-context extraction for AWS Lambda Durable Execution events.
The tracecontext propagation injection (i.e. checkpoint creation) part is done in [dd-trace-py PR#17773](https://github.com/DataDog/dd-trace-py/pull/17773)

Durable executions can resume with prior execution state instead of receiving the original trigger shape directly. This change teaches `extract_dd_trace_context` to recover Datadog propagation context from durable execution checkpoints, or from the original execution input payload when no checkpoint exists.

## Changes

- Detect durable execution events via `DurableExecutionArn`.
- Normalize `InitialExecutionState.Operations` whether it is provided as a list or operation map.
- Extract trace context from the latest `_datadog_{N}` checkpoint operation.
- Fall back to Datadog headers in the original `ExecutionDetails.InputPayload`.
- Preserve custom extractor precedence before durable-specific extraction.
- Continue through the existing extraction chain when no complete durable context is found.
- Add tests for:
  - latest checkpoint selection from an operation map
  - fallback extraction from JSON input payload headers

## Testing

- `python -m pytest -o addopts='' tests/test_tracing.py::TestExtractAndGetDDTraceContext::test_extracts_durable_trace_context_from_latest_checkpoint_operation_map tests/test_tracing.py::TestExtractAndGetDDTraceContext::test_extracts_durable_trace_context_from_input_payload_when_no_checkpoint`
  - 2 passed


datadog_lambda/tracing.py
```diff
@@ -546,6 +547,56 @@ def extract_context_from_step_functions(event, lambda_context):
         return extract_context_from_lambda_context(lambda_context)
 
 
+def _extract_context_from_durable_checkpoint(operation):
```
pablomartinezbernardo: nit: should these functions live in datadog_lambda/durable.py?

datadog_lambda/tracing.py
```diff
@@ -546,6 +547,56 @@ def extract_context_from_step_functions(event, lambda_context):
...
+            result = json.loads(result)
+        except Exception:
+            return None
+
+    if not isinstance(result, dict):
+        return None
+
+    # Checkpoints are written by dd-trace-py as x-datadog-* headers, so extract
+    # directly and bypass DD_TRACE_PROPAGATION_STYLE_EXTRACT on purpose.
+    return _DatadogMultiHeader._extract(result)
```
pablomartinezbernardo: Can this cause issues if tomorrow ddtrace refactors this class?

datadog_lambda/tracing.py
```diff
@@ -633,9 +684,12 @@ def extract_dd_trace_context(
     global dd_trace_context
     trace_context_source = None
     event_source = parse_event_source(event)
+    context = None
 
     if extractor is not None:
         context = extract_context_custom_extractor(extractor, event, lambda_context)
+    elif isinstance(event, dict) and "DurableExecutionArn" in event:
+        context = extract_context_from_durable_execution(event)
```
pablomartinezbernardo: What happens in the first invocation here? `isinstance(event, dict) and "DurableExecutionArn" in event` are both true, but `context` will be `None` because there is still no `_datadog_N` entry.

Do we expect the first invocation to extract the context with the other methods?

datadog_lambda/tracing.py
joeyzhao2018: updated this. this is actually unnecessary. good catch!

datadog_lambda/tracing.py
joeyzhao2018: moved

datadog_lambda/tracing.py
joeyzhao2018: Fixed. Changed the order of the extractors calls. http headers will now take precedence over the checkpoints

pablomartinezbernardo (approved):

datadog_lambda/durable.py
```diff
@@ -56,6 +59,62 @@ def extract_durable_function_tags(event):
 VALID_DURABLE_STATUSES = {"SUCCEEDED", "FAILED", "PENDING"}
 
 
+def _extract_context_from_durable_checkpoint(operation):
+    # Checkpoint data is written by the dd-trace-py in Datadog style
+    # (x-datadog-* headers). Extraction goes through the standard
+    # propagator.extract path, which honors DD_TRACE_PROPAGATION_STYLE_EXTRACT.
+    # The default extract list (datadog, tracecontext, baggage) already
+    # includes datadog. Customers who override the extract list MUST keep
+    # datadog in it.
```
pablomartinezbernardo, line 68: Should we document this somewhere? Do we have a workaround so compatibility doesn't depend on customer config?

joeyzhao2018: We will put this in the public documentation. the alternative was my previous commit which introduces some unnecessary complexity.

PR #824, opened 2026-05-08
build: fix dd-trace-py git overrides and filter publish layer check

joeyzhao2018:
# Summary

Two small build-script fixes I hit while trying to build a single layer from a dd-trace-py feature branch:

1. **`scripts/build_layers.sh`** — `DD_TRACE_COMMIT` and `DD_TRACE_COMMIT_BRANCH` were broken. Both paths build dd-trace-py from a git checkout, but the script wrote `ddtrace_serverless = { git = ... }` into `pyproject.toml`. Source checkouts publish project metadata as `ddtrace`, not `ddtrace_serverless` — the `ddtrace_serverless` variant only exists as pre-built wheels from dd-trace-py's CI (consumed via `UPSTREAM_PIPELINE_ID`). pip rejected the install with:

   > Requested ddtrace from git+...@<branch> has inconsistent name: expected 'ddtrace-serverless', but metadata has 'ddtrace'

   Fix: hard-code `ddtrace` as the package name in the two git-override branches. `DD_TRACE_WHEEL` and `UPSTREAM_PIPELINE_ID` paths are unchanged, so the release pipeline is unaffected.

2. **`scripts/publish_layers.sh`** — the pre-flight check looped over the full hard-coded `LAYER_PATHS` (14 entries) and bailed if any zip was missing, even when `LAYERS` was set to a single layer. So you couldn't publish one layer without first building all 14 zips. Moved the check to *after* `LAYERS` is resolved, and only verify files for layers we'll actually publish. When `LAYERS` is unset it still resolves to all available layers, so the default behavior (and the release pipeline that builds all layers) is unchanged.

## Backwards compatibility

- `DD_TRACE_WHEEL` / `UPSTREAM_PIPELINE_ID` / no-override default → unchanged.
- `LAYERS` unset → still requires all 14 zips, identical to before.
- Only relaxes one check and unbreaks two previously-broken env-var paths; nothing tightens.

## Test plan

- [x] `DD_TRACE_COMMIT_BRANCH=<branch> PYTHON_VERSION=12 ./scripts/build_layers.sh` produces `datadog_lambda_py-{amd64,arm64}-3.12.zip`.
- [x] `VERSION=N REGIONS=us-east-2 LAYERS=Datadog-Python312 ./scripts/publish_layers.sh` proceeds past the existence check with only the 3.12 amd64 zip built.
- [x] Existing release-pipeline invocations (`UPSTREAM_PIPELINE_ID=...`, no `LAYERS` filter) continue to work.


rithikanarayan:
I'm not sure if I'm mixing up two things here, but problem 1 might be getting solved by https://github.com/DataDog/dd-trace-py/pull/17964?

joeyzhao2018:
> I'm not sure if I'm mixing up two things here, but problem 1 might be getting solved by [DataDog/dd-trace-py#17964](https://github.com/DataDog/dd-trace-py/pull/17964)?

@rithikanarayan Yeah, probably. We can close this pr if that one get merged and working

PR #826, opened 2026-05-15
test: avoid side effects from mutating environ

KowalskiThomas:
### What does this PR do?

This changes the way some unit tests work to make sure running them does not pollute / change the `environ` for subsequent test runs.

We have noticed this is an issue in a recent incident where we were not able to set certain configuration flags for `ddtrace` because some tests would clear the environment, resulting in the configuration flags being removed and `ddtrace` thus using the default values for all flags.  
This made it harder to debug in the incident as one of the flags was causing crashes, and it was not possible to disable it.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -23,7 +23,7 @@ class TestDatadogLambdaAPI(unittest.TestCase):
         self.env_patcher.start()
 
     def tearDown(self):
-        del os.environ["AWS_REGION"]
+        self.env_patcher.stop()
 
     @patch("datadog_lambda.config.Config.fips_mode_enabled", True)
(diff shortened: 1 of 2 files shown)


rithikanarayan (changes requested):
The changes here look good to me. Can we also modify `test_profiling_enabled` in `test_wrapper.py` in this PR to stop the profiler at the end of that unit test. That would have this cover more of the mutated environment tests. If you disagree and think that should be a separate PR, I'm fine with that, too. Thanks!

KowalskiThomas:
@rithikanarayan should be good now, thanks for suggesting it!

rithikanarayan (approved):
:shipit:

PR #827, opened 2026-05-20
Remove update_deps.yml

pablomartinezbernardo:
### What does this PR do?

Removes the disabled `update-deps` workflow. The workflow was disabled in repo settings on january for failures and it has had no successful runs since 2025. Even before it broke, the workflow was effectively ignored.

### Motivation

Security noticed some secrets being unused, this removes all references to them from the repo


jcstorms1 (approved):
LGTM 👍

PR #828, opened 2026-05-20
Add unit tests to Ci Visibility

rithikanarayan:
### What does this PR do?
Adds use of CI Visibility tool in the unit tests.

### Motivation
Tracking our CI better (flaky tests, etc) in the [ddserverless org ](https://ddserverless.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40git.repository.id_v2%3Agithub.com%2Fdatadog%2Fdatadog-lambda-python&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&index=citest&start=1779295666192&end=1779900466192&paused=false).

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


ci/input_files/build.yaml.tpl
```diff
@@ -82,11 +82,19 @@ unit-test ({{ $runtime.name }}-{{ $runtime.arch }}):
   {{ end }}
   image: registry.ddbuild.io/images/mirror/python:{{ $runtime.image }}
   cache: &{{ $runtime.name }}-{{ $runtime.arch }}-cache
+  variables:
+    DD_CIVISIBILITY_AGENTLESS_ENABLED: "true"
+    DD_SERVICE: "datadog-lambda-python-{{ $runtime.python_version }}-{{ $runtime.arch }}"
```
duncanista: should runtime and arch be tags?

ci/input_files/build.yaml.tpl
```diff
@@ -82,11 +82,19 @@ unit-test ({{ $runtime.name }}-{{ $runtime.arch }}):
...
+    DD_ENV: "ci"
   before_script:
+    - source ./ci/get_dd_api_key.sh
```
duncanista: doesn't the credential script already do this?

ci/input_files/build.yaml.tpl
```diff
@@ -82,11 +82,19 @@ unit-test ({{ $runtime.name }}-{{ $runtime.arch }}):
...
     - PYTHON_VERSION={{ $runtime.python_version }} ARCH={{ $runtime.arch }} ./scripts/setup_python_env.sh
   script:
-    - source venv/bin/activate
-    - pytest -vv
+    - |
+      set -e
+      source ./ci/get_dd_api_key.sh
```
duncanista: sourcing the script again?

duncanista (approved):
approving to unblock, left a couple comments

tests/test_tracing.py
```diff
@@ -1045,13 +1045,15 @@ def test_set_correlation_ids(self):
...
+        # tracer.current_span() is None — pytest's CI Visibility plugin (--ddtrace)
+        # keeps a root test span active for the duration of the test.
         self.mock_get_dd_trace_context.return_value = Context()
-        # no exception thrown
-        set_correlation_ids()
-        span = tracer.current_span()
-        self.assertIsNone(span)
+        with patch.object(tracer, "trace") as mock_trace:
+            set_correlation_ids()
+            mock_trace.assert_not_called()
```
purple4reina, line 1056: Why is this change needed?

scripts/run_integration_tests.sh
```diff
@@ -220,6 +220,8 @@ for handler_name in "${LAMBDA_HANDLERS[@]}"; do
                 # Normalize python-requests version
                 sed -E "s/(User-Agent:python-requests\/)[0-9]+\.[0-9]+\.[0-9]+/\1X\.X\.X/g" |
                 sed -E "s/(\"http.useragent\"\: \"python-requests\/)[0-9]+\.[0-9]+\.[0-9]+/\1X\.X\.X/g" |
+                # ddtrace 4.x adds http.status_msg (derived from http.status_code); omit from snapshots
+                sed '/"http\.status_msg"/d' |
```
purple4reina: Don't we want to keep the status message? I'm assuming it's something like `200 OK`, which should match each time.

tests/test_tracing.py
rithikanarayan: With the CI Viz tool being used, it is no longer true that there is no current span. There is now the base span that ci viz uses to track all the tests. This new assertion still checks that when we set correlation ids with an empty context that we do not create a dummy span and instead correctly return early for set_correlation_ids, it just checks this in a different way.

scripts/run_integration_tests.sh
rithikanarayan: The status message is "OK" or "Accepted" and is related to but separate from the status code, which we still check. I could alternatively regenerate the snapshots to avoid doing this but since we have the status code tag already I thought removing it would be ok. Completely open to regenerating though.

purple4reina: Ah, I see.  ddtrace 4.x adds a new span tag, so this just deletes it.  Yeah, I think it would be best to just retake the snapshots.  I can imagine in the future it potentially being helpful to see when/if that tag disappears or changes.

tests/test_tracing.py
purple4reina: ahhh, makes sense!

purple4reina (approved):

PR #834, opened 2026-06-18
fix: use DD_SERVICE for inferred spans when integration service names are removed

zarirhamza:
## Summary

- When `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED=true` **and** `DD_SERVICE` is set, inferred (synthetic) spans use the base service name (`DD_SERVICE`) instead of the AWS resource/instance representation.
- An explicit `DD_SERVICE_MAPPING` entry still takes precedence over `DD_SERVICE`.
- Default behavior (flag off) is unchanged: inferred spans use the AWS resource name (queue/stream/table/etc.), so existing integration snapshots are unaffected.
- Adds `config.aws_service_representation_enabled` (`DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED`, default true) and `config.remove_integration_service_names_enabled` (`DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED`, default false).

## Resolution order in `determine_service_name`
1. `DD_SERVICE_MAPPING` (specific key, then generic key)
2. If `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` and `DD_SERVICE` set → `DD_SERVICE`
3. If AWS service representation disabled → fallback
4. Extracted resource/instance name, else fallback

## Motivation

Part of the Serverless Service Representation (SSR) work. The flag lets inferred spans collapse onto the function's base service when integration-specific service names are turned off, while keeping the resource-name behavior by default.

## Test plan

- [ ] `tests/test_tracing.py` (TestServiceMapping + inferred-span tests) pass with and without `DD_SERVICE` / the flag
- [ ] Integration snapshots unchanged (default behavior)


joeyzhao2018 (approved):
lgtm

PR #836, opened 2026-06-19
feat: add support for event bridge DSM context extraction

jeastham1993:
### What does this PR do?

Adds support for automatically extracting DSM context from Amazon Event Bridge events. Configures the DSM `exchange` tag using an environment variable.

### Motivation

Adding DSM support for all serverless messaging technologies.

### Testing Guidelines

Wrote unit test to test logical locally.
Compiled test version of layer and pushed to AWS account, tested with sample app.

### Types of Changes

- [ ] Bug fix
- [ X ] New feature
- [ ] Breaking change

### Check all that apply


datadog_lambda/tracing.py
```diff
@@ -82,6 +82,43 @@ def _dsm_set_checkpoint(context_json, event_type, arn):
...
+    """
+    if not config.data_streams_enabled:
+        return
+
+    if not detail_type:
+        return
+
+    try:
+        from ddtrace.internal.datastreams import data_streams_processor
+        from ddtrace.internal.datastreams.processor import PROPAGATION_KEY_BASE_64
```
purple4reina: We've had many problems in the past of using "internal" methods from ddtrace -- either ones in `internal` packages like this or ones starting with an underscore.

Are we certain that this api is stable? Do we have permission to use it? If someone tries to change it, is there a test that will fail over on the ddtrace side?

jeastham1993: Good question @purple4reina, let me ask the question

datadog_lambda/tracing.py
```diff
@@ -82,6 +82,43 @@ def _dsm_set_checkpoint(context_json, event_type, arn):
...
+
+        processor = data_streams_processor()
+        if not processor:
+            return
+
+        carrier_get = lambda k: context_json and context_json.get(k)  # noqa: E731
+        processor.decode_pathway_b64(carrier_get(PROPAGATION_KEY_BASE_64))
```
purple4reina: Why even define `carrier_get` if it's only used this once?

datadog_lambda/tracing.py
```diff
@@ -353,12 +390,32 @@ def _extract_context_from_eventbridge_sqs_event(event):
...
-    first_record = event.get("Records")[0]
+    records = event.get("Records")
+    first_record = records[0]
     body_str = first_record.get("body")
     body = json.loads(body_str)
     detail = body.get("detail")
+    # If `detail` is missing this is not an EventBridge -> SQS event; raising
+    # here lets the caller fall back to the regular SQS extraction path before
+    # any DSM checkpoint is set, avoiding double counting for plain SQS events.
     dd_context = detail.get("_datadog")
```
purple4reina: From my experience, python is really slow at raising/handling exceptions. Instead, can we just run some checks on `detail` and return early?

purple4reina: In the case where this is EventBridge -> SQS, should we create checkpoints for both? I'm not sure what the right thing to do would be.

datadog_lambda/tracing.py
```diff
@@ -353,12 +390,32 @@ def _extract_context_from_eventbridge_sqs_event(event):
     This is only possible if first record in `Records` contains a
     `body` field which contains the EventBridge `detail` as a JSON string.
     """
-    first_record = event.get("Records")[0]
+    records = event.get("Records")
+    first_record = records[0]
     body_str = first_record.get("body")
     body = json.loads(body_str)
```
purple4reina: We're json loading the first record twice. Seems inefficient.

purple4reina:
We should eventually add a new `data-streams` suite to the e2e tests.... eventually.

datadog_lambda/tracing.py
```diff
@@ -353,12 +390,32 @@ def _extract_context_from_eventbridge_sqs_event(event):
...
+    # checkpoint for every record in the batch. The message is consumed from
+    # the SQS queue, so it follows SQS conventions (type:sqs, topic:queue ARN).
+    if config.data_streams_enabled:
+        for record in records:
+            try:
+                record_body = json.loads(record.get("body"))
+                record_context = (record_body.get("detail") or {}).get("_datadog")
+                _dsm_set_checkpoint(
+                    record_context, "sqs", record.get("eventSourceARN", "")
+                )
```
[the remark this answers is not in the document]
purple4reina: Confirmed this is a real issue, not just an AI-review false positive, and needs to be addressed before merging.  We should also include a test to cover just this case to ensure it doesn't pop up again in the future.

purple4reina (changes requested):

datadog_lambda/tracing.py
robcarlan-datadog: I think ideally we would have a public API that allows supplying extra args such as `exchange`. But it doesn't exist right now. Practically it hasn't changed but there's nothing that would stop us from changing the signature because it isn't marked as a public api
Can we add an upstream change to enhance the public api? I'm not sure how the lambda versions are pinned to the python tracer

datadog_lambda/tracing.py
jeastham1993: @robcarlan-datadog what are your thoughts on this? My understanding is that if it's Event Bridge -> SQS we only emit the SQS checkpoint but because of the pathway hash that gets stitched together backend side?

datadog_lambda/tracing.py
jeastham1993: @purple4reina @robcarlan-datadog I'm happy to do the work to upstream the change to the Python API, but is this a blocker for merging this change?

purple4reina: Ideally, we wouldn't release any usage of an internal ddtrace api. But, as long as we make the changes in ddtrace then come back here to use those new public api's, I think we'd be fine. We just can't have usage of an internal api floating around without an ability to upgrade the layer past it.

tests/test_tracing.py
```diff
@@ -3646,3 +3647,188 @@ def test_kinesis_data_streams_disabled(self):
...
+
+        self.mock_checkpoint.assert_not_called()
+
+
+class TestEventBridgeDSMLogic(unittest.TestCase):
+    def setUp(self):
+        self.lambda_context = get_mock_context()
+        self.mock_processor = Mock()
+        tracer_patcher = patch(
+            "ddtrace.data_streams.ddtrace.tracer",
```
purple4reina: We should have at least one test that doesn't mock this api.  We need a way to detect if there's been a change in it upstream.

datadog_lambda/tracing.py
```diff
@@ -353,21 +395,52 @@ def _extract_context_from_eventbridge_sqs_event(event):
     This is only possible if first record in `Records` contains a
     `body` field which contains the EventBridge `detail` as a JSON string.
     """
-    first_record = event.get("Records")[0]
+    records = event.get("Records") or []
```
purple4reina: You can remove the `or []` since `not records` on the next line will capture the `None`. This avoids an empty allocation which can be slow in python.

tests/test_tracing.py
purple4reina: According to claude-code, this is important because python v3.10+ actually uses a different version of ddtrace which apparently doesn't contain the api's you're using.

datadog_lambda/tracing.py
```diff
@@ -82,6 +82,44 @@ def _dsm_set_checkpoint(context_json, event_type, arn):
...
+    """
+    if not config.data_streams_enabled:
+        return
+
+    if not detail_type:
+        return
+
+    try:
+        from ddtrace.data_streams import PROPAGATION_KEY_BASE_64
+        from ddtrace.data_streams import ddtrace as ddtrace_data_streams
```
purple4reina: There is no way this works. I checked `ddtrace.data_streams`, there is no `ddtrace` defined in that file.

We must have a test that exercises the new code paths without patching any of `ddtrace` to ensure that these api's exist and still work.

purple4reina (changes requested):

datadog_lambda/tracing.py
jeastham1993: @purple4reina for the moment I've reverted this back to use the exact same public API as SQS/SNS/Kinesis. I'll look at upstreaming the changes into dd-trace-py. We can hold off merging this PR until they have been merged and released

datadog_lambda/tracing.py
robcarlan-datadog: Yes this should just work as is!

purple4reina (approved):
Looks good.

There is a small bug here though, but I think it's okay as long as we call it out.

In the case where Eventbridge _and_ SNS get routed to the same SQS queue, then are batched into a single event payload, we'll miss the SNS dsm context.  This is because we search the entire batch for an eventbridge context and if found _anywhere_ in the batch, it will ignore SNS dsm extraction.

It's a problem, but I think very minor.  I think we should either add a comment about this or fix it if it's easy enough to address.

litianningdatadog:
You may want to update README with this new config `DD_DSM_EXCHANGE_NAME`.

datadog_lambda/tracing.py
```diff
@@ -353,22 +407,190 @@ def _extract_context_from_eventbridge_sqs_event(event):
...
+    """Return the ``_datadog`` carrier for a non-EventBridge SQS record.
+
+    A record in an SQS batch may itself be an SNS notification (an SNS => SQS
+    subscription without raw message delivery), in which case the attributes
+    live in the SNS envelope inside ``body`` rather than in the record's own
+    ``messageAttributes``. Both shapes are handled here so that a mixed batch
+    (EventBridge deliveries alongside SNS deliveries on the same queue) does
+    not lose the SNS DSM context.
+    """
+    msg_attributes = record.get("messageAttributes")
```
litianningdatadog: nit: what if messageAttributes is an empty object rather than being None?

litianningdatadog (commented):
left some minor comments

